PR #5, opened 2023-05-11
🤖 Agents: Chain of Thought (CoT) agent

andreibondarev:
Introducing Agents & Tools:

Agents are semi-autonomous bots that can respond to user questions and use available to them Tools to provide informed replies. They break down problems into series of steps and define Actions (and Action Inputs) along the way that are executed and fed back to them as additional information. Once an Agent decides that it has the Final Answer it responds with it.

Tools are (typically) APIs to various 3rd party services that Agents that use to execute Actions with.

```ruby
agent = Agent::ChainOfThoughtAgent.new(llm: :openai, llm_api_key: ENV["OPENAI_API_KEY"], tools: ['search', 'calculator'])
agent.run(question: "How many full soccer fields would be needed to cover the distance between NYC and DC in a straight line?", logging: true)
```


rickychilcott:
This is awesome stuff. Can't wait to dig in @andreibondarev!

PR #21, opened 2023-05-19
Discontinuing the use of Weaviate modules

moekidev:
ref: https://github.com/andreibondarev/weaviate-ruby/pull/18#issuecomment-1554318857

I think we should embed via Langchain.rb and then passing it to Weaviate when integrating Langchain.rb and Weaviate.

https://weaviate.io/developers/weaviate/api/rest/objects#with-a-custom-vector

That way, we can align with other vector databases and APIs.
Furthermore, for Langchain users, the benefits of the Weaviate module are not that substantial.


andreibondarev:
@moekidev Do you think this produces better results? By not relying on the Weaviate internal modules.

moekidev:
For maintainability. It helps reduce [accidental reuse of LLM names](https://github.com/andreibondarev/langchainrb/blob/5796dc0c9bfcb2514f6ae2d6f5b3cea33bd486b3/lib/vectorsearch/weaviate.rb#L46) and [branching based on the presence or absence of modules](https://github.com/andreibondarev/langchainrb/blob/5796dc0c9bfcb2514f6ae2d6f5b3cea33bd486b3/lib/vectorsearch/weaviate.rb#L97-L117). Of course, it is possible to rely on the extensibility of Weaviate modules. However, currently, Langchain.rb is limited to OpenAI, Cohere, and HuggingFace APIs to align with other vector databases. As a result, it does not benefit from the full extent of extensibility. Nevertheless, I believe this is the right choice. It is advisable to have Langchain.rb designed in a way that allows for easy replacement of the database and language model, ensuring compatibility among APIs.

andreibondarev:
@moekidev Could you please run `standardrb --fix` and commit to fix the failing specs?

lib/vectorsearch/weaviate.rb
```diff
@@ -63,14 +58,9 @@ def create_default_schema
     # @param k [Integer|String] The number of results to return
     # @return [Hash] The search results
     def similarity_search(query:, k: 4)
-      near_text = "{ concepts: [\"#{query}\"] }"
+      embedding = generate_embedding(text: query)
```
andreibondarev, line 61: Right now we just use `llm_client.embed`. `generate_embedding` alias was removed.

spec/vectorsearch/weaviate_spec.rb
```diff
@@ -55,11 +70,25 @@
...
+          {
+            "embedding" => [
+              -0.0018150936,
+              0.0017554426,
+              -0.022715086
+            ]
+          }
+        ]
+      })
     end
```
andreibondarev, line 93: Also I would really appreciate if you can write the spec for the `#ask` method. Thank you!

andreibondarev:
@moekidev A few comments and it's good to go! 🚀

spec/vectorsearch/weaviate_spec.rb
moekidev: OK. I wrote!

moekidev:
@andreibondarev Please check.

moekidev:
@andreibondarev Could you please review it again?

PR #43, opened 2023-05-22
Escape double curly braces in prompt template strings

### LINKED ISSUE #42: Support for JSON examples in PromptTemplate
It's common when prompt engineering to require the LLM to return results in JSON format and to include an example JSON response within the prompt. The current Prompt variable parser can't handle this as it interprets the example JSON as an input variable and then complains that it is missing:

```ruby
require "langchain"
simple_template = "Tell me a {adjective} joke. Return in JSON in the format {{joke: 'The joke'}}"
Prompt::Base.extract_variables_from_template(simple_template)
=> ["adjective", "joke: 'The joke'"]
```

The current Python f-string parser handles this by allowing you to escape a curly brace with a double curly brace, as shown in the example above. [Python f-string spec](https://docs.python.org/3/reference/lexical_analysis.html#f-strings)

Looking at the regex its clearly trying to do something with double curly brackets so before I try and fix it can you shed some light on the original implementation?

timrwilliams:
Resolves #42 

---
Simple approach to not include input variables denoted by double curly braces. As per the f-string syntax double curly braces are replaces by single curly braces in the output.

```ruby
prompt = Prompt::PromptTemplate.new(template: "Tell me a {adjective} joke. Return in JSON in the format {{joke: 'The joke'}}", input_variables: ["adjective"])
prompt.format
=> "Tell me a {adjective} joke. Return in JSON in the format {joke: 'The joke'}"
irb(main):006:0>
```

Tested via new spec and also ran against all current Prompt examples on langchain-hub to ensure compatibility with wider ecosystem.

### DIFF
--- a/lib/prompt/base.rb
+++ b/lib/prompt/base.rb
@@ -74,9 +74,9 @@ module Prompt
       input_variables = []
       scanner = StringScanner.new(template)
 
-      while scanner.scan_until(/\{([^{}]*)\}/)
+      while scanner.scan_until(/\{([^}]*)\}/)
         variable = scanner[1].strip
-        input_variables << variable unless variable.empty?
+        input_variables << variable unless variable.empty? || variable[0] == "{"
       end
 
       input_variables
--- a/lib/prompt/base.rb
+++ b/lib/prompt/base.rb
@@ -66,6 +66,8 @@ module Prompt
     # contained within the template. Input variables are defined as text enclosed in
     # curly braces (e.g. "{variable_name}").
     #
+    # Content within two consecutive curly braces (e.g. "{{ignore_me}}) are ignored.
+    #
     # @param template [String] The template string to extract variables from.
     #
     # @return [Array<String>] An array of input variable names.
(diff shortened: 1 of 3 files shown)


lib/prompt/base.rb
```diff
@@ -74,9 +74,9 @@ def self.extract_variables_from_template(template)
       input_variables = []
       scanner = StringScanner.new(template)
 
-      while scanner.scan_until(/\{([^{}]*)\}/)
+      while scanner.scan_until(/\{([^}]*)\}/)
         variable = scanner[1].strip
-        input_variables << variable unless variable.empty?
+        input_variables << variable unless variable.empty? || variable[0] == "{"
```
andreibondarev, line 79: Nitpicking: `variable.start_with?("{")` to improve the readability perhaps.

andreibondarev:
@timrwilliams Just 1 comment. 

Could you please upmerge your branch and also fix the failing specs? I think it's just the `standardrb` check failing.

Also -- would you please add some code comments referencing the intention behind this fix? Thank you!

andreibondarev (approved):
@timrwilliams Looks great, thank you!!

PR #45, opened 2023-05-23
Loaders overhaul

alchaplinsky:
## Motivation
- Give developers **simple API** to load different types of data. Offload decision on which loader to use from developers as library is able to find the right processor based on extension or content type.
- Support both **local file path** and **URL** as a source of data.
- Make it **easier to add more data types** like (JSON, JSONL, CSV, etc.). No need to register loaders/processors. The only thing needed is just adding a new processor to Langchain::Processors namespace:
```ruby
class Langchain::Processors::Custom
   EXTENSIONS = []
   CONTENT_TYPES = []

   def parse(data)
     # custom parsing logic
   end
end

```

## Changes
- Make `Langchain::Loader` smart and allow it determine how to load data and which processor to use for downloaded data.
- Remove Loader classes like `Loaders::Text`, `Loaders::PDF`, etc.
- Add Processor classes which responsibility is to define `parse` function and `EXTENSIONS` and `CONTENT_TYPES` constants.
- Refactor `Vectorsearch::Base` to use new `Loaders::Base` loader instead of global `Loader` class. No need to call `add_loader` now before adding data.
- Remove `Loader` class. Seems like it is not used anywhere else.
- Specs updates 

## API
Loader can be used by either instantiating a class or calling `load` method directly on the loader class.

```ruby
loader = Langchain::Loader.new(path)
loader.load
```

or 
```ruby
Langchain::Loader.load(path)
```

Path parameter can be either a local file path or a URL:
```ruby
Langchain::Loader.load('spec/fixtures/loaders/example.html')
```

or

```ruby
Langchain::Loader.load('https://en.wikipedia.org/wiki/Guardians_of_the_Galaxy_Vol._3')
```

## ToDo
- [x] Decide on naming: `Processors` vs `Parsers`, `Loaders::Base` vs `Langchain::Loader` or any other naming concerns/suggestions?
- [ ] Allow passing a custom processor. Developers might want to pass a custom `parse` function if the data format is not yet supported by library or there's some special handling needs to be done.


rickychilcott:
This is all looking good to me and I appreciate the work you've done on it.

I don't have a strong preference on naming, though it does seem to me that these are `Parsers` now and not `Loaders`. `Loader` meaning it's responsible for preparing the file, url, pipe, stream, whatever and `Parser` for transforming that io object into meaningful data.

I also think we should be namespacing ALL classes/modules with Langchain so that it would be `Langchain::Vectorstore...`, `Langchain::Parsers...` because we'll get name collisions soon...

The only problem with not "registering" or having a controllable load order is that a custom parser that overlaps with criteria (i.e. my custom parser _might_ be able to parse a special form of HTML) doesn't give an exit hatch. So I'd be supportive of allowing you to specify the order -- at least eventually.

andreibondarev:
@alchaplinsky Tons of great work, thank you Alex!

@rickychilcott @alchaplinsky Let's finalize this (cleanup, specs, rdocs, etc.). We'll keep the "Processors" name for now.

The next thing we need to figure out is the Response object so instead of the string is an object that contains the metadata, to_s, etc.

And then I think we'll be able to "connect the dots" to Chunkers.

Does this make sense? You guys agree?

rickychilcott:
> @alchaplinsky Tons of great work, thank you Alex!
> 
> @rickychilcott @alchaplinsky Let's finalize this (cleanup, specs, rdocs, etc.). We'll keep the "Processors" name for now.
> 
> The next thing we need to figure out is the Response object so instead of the string is an object that contains the metadata, to_s, etc.
> 
> And then I think we'll be able to "connect the dots" to Chunkers.
> 
> Does this make sense? You guys agree?

Yes. Agreed!

@alchaplinsky - let me know if I can assist in any way.

I think we were planning to call this a `Result` object, but `ProcessedResult` might be better

```ruby
class ProcessedResult
  def initialize(**args)
    @metadata, @content = args[:metadata], args[:content]
  end

  attr_reader :metadata, :content
  delegate :to_s, to: :content
end
```

I'm not sure what the shape of `content` is, but we'll figure that out.

alchaplinsky:
@rickychilcott Once this one is merged I'll add a couple of other processors for common data types. And based on that, I plan to make a PR to update all these processors to return the Result object instead of string content. Your example above is a good start, we can extend it in the future as needed.

rickychilcott:
> @rickychilcott Once this one is merged I'll add a couple of other processors for common data types. And based on that, I plan to make a PR to update all these processors to return the Result object instead of string content. Your example above is a good start, we can extend it in the future as needed.

BTW - it appears that in Langchain, they emit a `Document`

```python
from langchain.docstore.document import Document

docs = [Document(page_content=t) for t in texts[:3]]
```

And this is similar to what I've seen in https://github.com/BoxcarsAI/boxcars/blob/main/lib/boxcars/vector_store/document.rb

So we may want to consider calling it `Document` instead of `Result` or `ProcessedResult`

alchaplinsky:
@rickychilcott Yeah, I've seen that. `Document` seems like a great name for PDF/Docx or event HTML and pages also make sense within the context of a document. However, for media like Audio or Video `Document` would probably not be that good.

andreibondarev:
I think a case could be made for an "audio document" or "video document". 

I'd rather start as lean as possible. We just need to be flexible, even if it's just a simple Struct. We will learn more as we go because it's not yet crystal clear what the common attributes should be uniting the different types of docs.

Thought of using Struct as the return value for the time being? Until we're confident and commit to a class.

alchaplinsky:
@andreibondarev I'll give it more thought. In the meantime, It's fine to return a string for now.

PR #46, opened 2023-05-23
Add namespace support for Pinecone client calls

faustus7:
**Motivation**

Add support for working with Pinecone vector database namespaces.  This would allow for finer-grained operations on namespaced data within the vector database index.

**Changes**

Updated `Vectorsearch::Pinecone` to accept and forward optional `namespace:` parameters across client methods.

**API**

Updated APIs in `Vectorsearch::Pinecone
* `def add_texts(texts:, namespace: '')`
* `def similarity_search(query:, k: 4, namespace: '')`
* `def similarity_search_by_vector(embedding:, k: 4, namespace: '')`
* `def ask(question:, namespace: '')`


andreibondarev:
@faustus7 Is it a better approach than passing `namespace:` in the initializer and setting it to an instance variable `@namespace = namespace` like we do with `@index_name`?

andreibondarev:
@faustus7 I'm just curious, I don't really have a strong preference!

faustus7:
> @faustus7 Is it a better approach than passing `namespace:` in the initializer and setting it to an instance variable `@namespace = namespace` like we do with `@index_name`?

I was thinking the same thing after opening the pull that it could be cleaner, but then I thought it might be better to have finer control over switching namespaces with searches after creating the client. Thoughts?

andreibondarev:
@ScotterC Would you happen to have any thoughts on this? Sorry to bug ya! :)

lib/vectorsearch/pinecone.rb
```diff
@@ -25,12 +25,14 @@ def initialize(environment:, api_key:, index_name:, llm:, llm_api_key:)
...
     # @param texts [Array] The list of texts to add
+    # @param namespace [String] The namespace to add the texts to
     # @return [Hash] The response from the server
-    def add_texts(texts:)
+    def add_texts(texts:, namespace: "")
       vectors = texts.map do |text|
         {
           # TODO: Allows passing in your own IDs
           id: SecureRandom.uuid,
+          namespace: namespace,
```
ScotterC, line 35: Important note. The readme on the Pinecone client was out of date. The namespace needs to be outside the vectors array and passed as a named argument to the upsert method

```ruby
index = client.index(index_name)

index.upsert(
  namespace: namespace,
  vectors: vectors
)
```

faustus7: Nice catch, thanks! https://docs.pinecone.io/reference/upsert

andreibondarev (approved):
@faustus7 Thank you, this is great!

andreibondarev (approved):
@faustus7 Thank you, this is great!

PR #48, opened 2023-05-23
Add require entrypoint to follow bundler convention

technicalpickles:
By default, bundler will try to require the name of the rubygem. For
example, this gem is langchainrb, so it would try to
`require "langchainrb"`.

That file doesn't exist, so when using bundler you'd need to explicitly
set `require`, ie:

```ruby
gem "langchainrb", require: "langchain"
```

This adds `lib/langchainrb.rb` that only requires "langchain" so that
extra `require: "langchain"` isn't
necessary.


rickychilcott (approved):
Lovely!

andreibondarev:
@technicalpickles Please run `standardrb --fix` and let's ship it! Thank you!

andreibondarev (approved):

technicalpickles:
Fixed!

PR #56, opened 2023-05-24
Add custom metadata and filter support for Pinecone client calls

faustus7:
**Motivation**

Add support for supplying custom metadata when upserting texts to a Pinecone vector index. This would allow for finer-grained filtration operations on queried data within the vector database index.

**Changes**

Updated Vectorsearch::Pinecone.add_texts to accept and forward an optional metadata: parameter.

**API**

Updated APIs in `Vectorsearch::Pinecone`

* `def add_texts(texts:, namespace: '', metadata: nil)`
* `def similarity_search(query:, k:, namespace: '', filter: nil)`
* `def similarity_search_by_vector(query:, k:, namespace: '', filter: nil)`
* `def ask(question:, namespace: '', filter: nil)`

### DIFF
--- a/lib/vectorsearch/pinecone.rb
+++ b/lib/vectorsearch/pinecone.rb
@@ -57,18 +57,21 @@ module Vectorsearch
     # @param query [String] The text to search for
     # @param k [Integer] The number of results to return
     # @param namespace [String] The namespace to search in
+    # @param filter [String] The filter to use
     # @return [Array] The list of results
     def similarity_search(
       query:,
       k: 4,
-      namespace: ""
+      namespace: "",
+      filter: nil
     )
       embedding = llm_client.embed(text: query)
 
       similarity_search_by_vector(
         embedding: embedding,
         k: k,
-        namespace: namespace
+        namespace: namespace,
+        filter: filter
       )
     end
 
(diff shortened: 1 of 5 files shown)


lib/vectorsearch/pinecone.rb
```diff
@@ -26,13 +26,14 @@ def initialize(environment:, api_key:, index_name:, llm:, llm_api_key:)
...
+    # @param metadata [Hash] The metadata to use for the texts
     # @return [Hash] The response from the server
-    def add_texts(texts:, namespace: "")
+    def add_texts(texts:, namespace: "", metadata: nil)
       vectors = texts.map do |text|
         {
           # TODO: Allows passing in your own IDs
           id: SecureRandom.uuid,
-          metadata: {content: text},
+          metadata: metadata || {content: text},
```
andreibondarev, line 36: I can't recall why `{content: text}` was previously passed in. Is it a required attribute?

lib/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Langchain
-  VERSION = "0.3.12"
+  VERSION = "0.3.13"
```
andreibondarev, line 4: No need to update the version, I update it whenever I'm actually pushing out a new version.

lib/vectorsearch/pinecone.rb
faustus7: Based on Pinecone documentation, it seems like this can be any key:value metadata that we want to use to provide extra context around upserted vector values, and seemingly `content` is not a required or 'magic' key.

https://docs.pinecone.io/docs/insert-data#inserting-vectors-with-metadata 
https://docs.pinecone.io/docs/metadata-filtering
https://docs.pinecone.io/reference/upsert

andreibondarev (approved):

PR #68, opened 2023-05-26
Namespace

technicalpickles:
This should address https://github.com/andreibondarev/langchainrb/issues/47 . I did it commit by commit, so specs gave me some confidence things are working as expected.

I was not sure how y'all want to handle breaking changes like constants moving, so figured preserving that just for a bit would be preferable. I've done that by making top level constants that just point to the newly namedspaced ones. That means it doesn't address one of the concerns, ie namespace collisions, but it gets the rest of the code ready to change it.

Because that existing behavior is saved, I didn't update the existing README.


andreibondarev:
@technicalpickles Please run `standardrb --fix` and commit.

Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    langchainrb (0.3.13)
+    langchainrb (0.3.10)
```
andreibondarev, line 4: I think this change was accidental?

lib/langchain.rb
```diff
@@ -28,43 +42,35 @@ module Processors
...
-  autoload :Base, "tool/base"
-  autoload :Calculator, "tool/calculator"
-  autoload :SerpApi, "tool/serp_api"
-  autoload :Wikipedia, "tool/wikipedia"
-end
+Agent = Langchain::Agent
+LLM = Langchain::LLM
+Prompt = Langchain::Prompt
+Tool = Langchain::Tool
+Vectorsearch = Langchain::Vectorsearch
```
andreibondarev, line 76: I don't think we need these constants. I would just have to bump the version from 0.3.x to 0.4.x to indicate the breaking changes.

Gemfile.lock
rickychilcott: I noticed there was another 0.3.13 -> 0.3.10 in this same PR. Make sure you fix them both @technicalpickles.

Thanks for doing this work!

technicalpickles: Not sure what happened here. Fixed it in any event.

technicalpickles:
I've made the fixes, but looks like I have merge conflicts as I feared 😦 

I could try doing things in smaller batches, and use legacy namespaces to help keep things from conflicting as much.

andreibondarev:
> I've made the fixes, but looks like I have merge conflicts as I feared 😦
> 
> I could try doing things in smaller batches, and use legacy namespaces to help keep things from conflicting as much.

@technicalpickles Sorry about that! Why won't we try moving it over folder by folder? I'll be around to quickly merge things today as well.

I'd just start with these folders and go down the list: https://github.com/andreibondarev/langchainrb/tree/main/lib

PR #77, opened 2023-05-29
Allow passing custom processor as a block to a loader

alchaplinsky:
## Motivation
1. Loaders need the ability to pass options to processors. Example: CSV processor works with coma-separated columns by default. But fails to parse semicolons correctly. A way to pass a `col_sep` option to `CSV.new` within a processor should exist.
2. If no processor for a specific file format exists there's currently no way to load and process this file. By passing a block to `load` method user should be able specify a custom processor for a particular load call.

## Changes
1. Allow `Langchain::Loader` to accept both `options` hash and a block. A call can now look like this:

**Passing an option**:
```ruby
Langchain::Loader.load(path, { col_sep: ';' })
```
or
```ruby
Langchain::Loader.new(path, { col_sep: ';').load
```
**Passing a block**

```ruby
Langchain::Loader.load(path) do |data|
  run_custom_processing(data)
end
```
or 

```ruby
Langchain::Loader.new(path).load do |data|
  run_custom_processing(data)
end
```
2. Accept `col_sep` option within CSV processor so that different separators are accepted within `csv` files

### DIFF
--- a/lib/langchain/loader.rb
+++ b/lib/langchain/loader.rb
@@ -39,10 +39,10 @@ module Langchain
     def load(&block)
       @raw_data = url? ? load_from_url : load_from_path
 
-      if block_given?
-        data = yield @raw_data.read, @options
+      data = if block
+        yield @raw_data.read, @options
       else
-        data = processor_klass.new(@options).parse(@raw_data)
+        processor_klass.new(@options).parse(@raw_data)
       end
 
       Langchain::Data.new(data, source: @path)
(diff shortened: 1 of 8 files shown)


andreibondarev (commented):
@alchaplinsky Looks good.

Is the idea here that some of these passed in options will be passed all the way down to the Chunkers as well?

@rickychilcott All good with you?

alchaplinsky:
@andreibondarev I mean we can pass any of these options to any modules down the line. However, since we decided to make chunkers decoupled from loaders it makes sense to pass options to chunker directly.

andreibondarev (approved):

PR #88, opened 2023-06-01
Code interpreter tool

bborn:
First try at adding a very simple code interpreter tool using `safe_ruby` to prevent malicious code execution.


lib/langchain.rb
```diff
@@ -73,4 +73,5 @@ module Tool
   autoload :Calculator, "tool/calculator"
   autoload :SerpApi, "tool/serp_api"
   autoload :Wikipedia, "tool/wikipedia"
+  autoload :CodeInterpreter, "tool/code_interpreter"
```
andreibondarev, line 76: Please move to under "Calculator" (alphabetical order)

lib/tool/code_interpreter.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Tool
+  class CodeInterpreter < Base
+    description <<~DESC
+      Useful for executing ruby code.
```
andreibondarev, line 6: Is this the best description we could provide it? I'm curious if anyone else has done it and what kind of description have they used. Like should we tell it that the expected input to this tool is a valid Ruby code or expression? May need to be tested out. #promptengineering.

spec/tool/code_interpreter_spec.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
```
andreibondarev, line 2: I think you may need `require "safe_ruby"` but maybe not.

lib/tool/code_interpreter.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Tool
+  class CodeInterpreter < Base
+    description <<~DESC
```
andreibondarev, line 5: Would you please write a description to this Tool (code comments) like what this tool does, how to use it, etc. Thank you!

lib/tool/code_interpreter.rb
bborn: Good point. How about this (based on LC):

```A Ruby code interpreter. Use this to execute ruby expressions. Input should be a valid ruby expression. If you want to see the output of the tool, make sure to return a value.```

File: spec/tool/code_interpreter_spec.rb (hunk 3 above)
Comment (bborn) [reply spec/tool/code_interpreter_spec.rb:2]: Seems to run fine without it.

**Comment by andreibondarev** [comment]
@bborn You need to add the tool to this list: https://github.com/andreibondarev/langchainrb/blob/main/lib/tool/base.rb#L12-L16

File: lib/tool/code_interpreter.rb [hunk 5]
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Tool
+  class CodeInterpreter < Base
```
rickychilcott, line 4: Do you think it makes sense to call this `RubyCodeInterpreter`? Of course we won't be able to execute many languages, but I could see us writing a Shell, JS, or other kinds of code interpreters.

andreibondarev:
@bborn Have you had much time testing it? Here's one test I tried:

```
irb(main):009:0> agent.run(question: "Use a Ruby program to output the sum of Fibonacci sequence for 1 to 1000.")
[LangChain.rb] Agent: Passing the prompt to the openai LLM
[LangChain.rb] Agent: Using the "code_interpreter" Tool with "puts (1..1000).inject(0) { |sum, n| sum + (n < 2 ? n : (n-1) + (n-2)) }"
Traceback (most recent call last):
       12: from bin/console:60:in `<main>'
       11: from (irb):6:in `<main>'
       10: from (irb):7:in `rescue in <main>'
        9: from /Users/andrei/Code/langchain/lib/agent/chain_of_thought_agent/chain_of_thought_agent.rb:44:in `run'
        8: from /Users/andrei/Code/langchain/lib/agent/chain_of_thought_agent/chain_of_thought_agent.rb:44:in `loop'
        7: from /Users/andrei/Code/langchain/lib/agent/chain_of_thought_agent/chain_of_thought_agent.rb:67:in `block in run'
        6: from /Users/andrei/Code/langchain/lib/tool/base.rb:27:in `execute'
        5: from /Users/andrei/Code/langchain/lib/tool/code_interpreter.rb:18:in `execute'
        4: from /Users/andrei/Code/langchain/lib/tool/code_interpreter.rb:22:in `safe_eval'
        3: from /Users/andrei/.asdf/installs/ruby/3.0.0/lib/ruby/gems/3.0.0/gems/safe_ruby-1.0.4/lib/safe_ruby/runner.rb:20:in `eval'
        2: from /Users/andrei/.asdf/installs/ruby/3.0.0/lib/ruby/gems/3.0.0/gems/safe_ruby-1.0.4/lib/safe_ruby/runner.rb:41:in `eval'
        1: from /Users/andrei/.asdf/installs/ruby/3.0.0/lib/ruby/gems/3.0.0/gems/safe_ruby-1.0.4/lib/safe_ruby/runner.rb:45:in `rescue in eval'
RuntimeError (998002)
0
```
Throws an error at the end ^^

In some other instances it tried to execute something like:
```
require "fibonacci"
```

There may be additional instruction we may need to give it like "use only the standard Ruby library" and "you can't require any outside libraries" or "you only have access to the Ruby runtime". I'm not sure how we'd test all of this.

andreibondarev:
@bborn Also the [safe_ruby gem](https://github.com/ukutaht/safe_ruby) itself seems to be super old and abandoned. Are there any other better options?

bborn:
Yeah ... not much beyond what's in the spec. I think `SafeRuby` raises an error with `puts`. Agreed that the tool description probably needs to describe some of the more obvious unalloyed things (like `require`)

Agreed about SafeRuby, it was just the first thing I found that works. 

https://github.com/tario/shikashi - this looks better maintained (and a bit nicer), but it requires whitelisting all the allowed methods/classes. I guess SafeRuby does the [same here](https://github.com/ukutaht/safe_ruby/blob/master/lib/method_whitelist.rb) so maybe it would be easy to just steal that whitelist.

andreibondarev:
@bborn I'm not opposed to `safe_ruby`, we can always fork and update it. Just need to figure out why `puts` throws an error and find any other issues.

lib/tool/code_interpreter.rb
bborn: @rickychilcott yes, I think that's a good point.

bborn:
@andreibondarev `puts` throws an error because it's not whitelisted here: https://github.com/ukutaht/safe_ruby/blob/master/lib/method_whitelist.rb

andreibondarev:
@bborn You also need to resolve merge conflicts.

PR #90, opened 2023-06-01
Switch the default operator to cosine distance for pgvector

### LINKED ISSUE #86: Pgvector integration `similarity_search_by_vector()` should use the cosine distance by default
The Pgvector's `similarity_search_by_vector()` should be calculating the cosine distance by default: https://github.com/andreibondarev/langchainrb/blob/main/lib/vectorsearch/pgvector.rb#L75-L77

More info about building a cosine distance query here: https://github.com/pgvector/pgvector/blob/dee2c4feb1bc5b17b9fe6a0a1ce8dbf0963c1b05/README.md#vector-operators

ProGM:
Fixes #86

I've switched the operator to cosine distance (<=>) and added a test.

I've put the operators into a constant with a proper naming, since `<=>` is not very explicit.

I'm wondering if we want to give the option to change the operator, maybe in the **constructor**?

Something like:

```ruby
vector_store = Vectorsearch::Pgvector.new(url: ENV['POSTGRES_URL'], operator: 'euclidean_distance')
...
```

### DIFF
--- a/lib/vectorsearch/pgvector.rb
+++ b/lib/vectorsearch/pgvector.rb
@@ -3,6 +3,13 @@
 module Vectorsearch
   # The PostgreSQL vector search adapter
   class Pgvector < Base
+    # The operators supported by the PostgreSQL vector search adapter
+    OPERATORS = {
+      "cosine_distance" => "<=>",
+      "euclidean_distance" => "<->"
+    }
+    DEFAULT_OPERATOR = "cosine_distance"
+
     # @param url [String] The URL of the PostgreSQL database
     # @param index_name [String] The name of the table to use for the index
     # @param llm [String] The URL of the Language Layer API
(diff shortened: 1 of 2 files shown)


andreibondarev (approved):

andreibondarev:
> I'm wondering if we want to give the option to change the operator, maybe in the constructor?

@ProGM Not a bad idea but we can probably hold off for now.

lib/vectorsearch/pgvector.rb
```diff
@@ -73,7 +81,7 @@ def similarity_search_by_vector(embedding:, k: 4)
       result = client.transaction do |conn|
         conn.exec("SET LOCAL ivfflat.probes = 10;")
         query = <<~SQL
-          SELECT id, content FROM #{@index_name} ORDER BY vectors <-> $1 ASC LIMIT $2;
+          SELECT id, content FROM #{@index_name} ORDER BY vectors #{@operator} $1 ASC LIMIT $2;
```
andreibondarev, line 84: We should probably add:
```
attr_reader :operator
```
at the top of the class, so we're not accessing the instance variable directly.

Issue #80, opened 2023-05-30
An abstraction that interfaces with a database and uses an LLM to come up with SQL queries to be executed.


andreibondarev:
## Background
A popular use-case is constructing a prompt for the LLM that includes the database schema, and user's question at hand to construct a SQL query that gets executed on LLM's behalf. The SQL query's result then gets fed back to the LLM to synthesize an answer to the user's question.

 ## Objectives
Create an entity (should it be a Tool? an Agent? or another abstraction?) that:
- Can establish a connection to a database
- Can execute (sanitized!) SQL queries
- Can construct a prompt for the LLM that:
  - Includes information about the DB schema, or only relevant tables? Subject to context window limits.
  - Includes the instructions for the LLM
  - Includes the user's question.
  - Good prompt to model after: [link](https://github.com/jerryjliu/llama_index/blob/35aa5bee71a6f7ef6e1c331a2f7528dfc52b7ffc/llama_index/prompts/default_prompts.py#L171-L192).

ZhenhangTung:
@andreibondarev I'll try to rephrase the challenge to make sure I understand it correctly. What we want to build is an abstraction that take user's question as input and return LLM's response as output. During this process, we will connect to the db searching texts that share similar semantic meaning and use them as prompt's context. If that's the case, we could consider implementing [RetrevalQA](https://blog.langchain.dev/retrieval/)?

PR #91, opened 2023-06-01
Initial SQLQueryAgent implementation

mattlindsey:
Work in progress PR to see if there are any comments.
For #80 

I still need to hook in Sequel calls and see how well it works, and add testing.
Also I should get and set the SQL dialect.
And review safety concerns here: https://github.com/jeremyevans/sequel/blob/master/doc/security.rdoc


andreibondarev:
@mattlindsey Looks like it's off to a great start! You'll have to upmerge, I did a rather large namespace change across the whole project.

andreibondarev:
Given your work here, we may need to expand how the Agents and Tools are initialized. 

Previously:
```ruby
agent = Langchain::Agent::ChainOfThoughtAgent.new(llm: :openai, llm_api_key: ENV["OPENAI_API_KEY"], tools: ['search', 'calculator', 'database'])
```
will probably need to change to accept options for Tools. E.g.:
```ruby
agent = Langchain::Agent::ChainOfThoughtAgent.new(llm: :openai, llm_api_key: ENV["OPENAI_API_KEY"], tools: ['search', 'calculator', 'database'], database_options: {}, search_options: {}, calculator_options: {})
```
OR
```ruby
calculator = Langchain::Tool::Calculator.new(...)
database  = Langchain::Tool::Database.new(...)

agent = Langchain::Agent::ChainOfThoughtAgent.new(llm: :openai, llm_api_key: ENV["OPENAI_API_KEY"], tools: [calculator, database]
```

lib/tool/database.rb
```diff
@@ -0,0 +1,29 @@
...
+      The input to this tool should be valid SQL.
+    DESC
+
+    def initialize(db_connection_string)
+      depends_on "sequel"
+      require "sequel"
+    end
+
+    def schema
+      return "the_schema_string"
```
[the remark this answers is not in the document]
bborn: See: https://arxiv.org/abs/2204.00498 for some research.

I like the way LlamaIndex does it here:

https://github.com/jerryjliu/llama_index/blob/b4618a2a24cd11b5c5949ab97389d62ac34ea336/llama_index/indices/struct_store/container_builder.py#L76

They just pull the table names and info and stuff them in the prompt as a string. 

When the number of tables is too big, this won't work (context window), so they dump the table schema into a vector index and query that for the relevant context:

https://gpt-index.readthedocs.io/en/latest/examples/index_structs/struct_indices/SQLIndexDemo-ManyTables.html#using-llamaindex-to-store-table-schema-context

lib/agent/sql_query_agent/sql_query_agent_answer_prompt.json
```diff
@@ -0,0 +1,9 @@
+{
```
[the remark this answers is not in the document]
mattlindsey: I don't mind the JSON, but whatever you think is fine.

lib/tool/database.rb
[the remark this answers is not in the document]
mattlindsey: Thanks @bborn. I'll look at that.  I added code to get the schema but a large schema currently blows up like you said.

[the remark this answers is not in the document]
andreibondarev: @bborn Super helpful, thank you! I don't think we'll get all of the edge-cases addresses in this first iteration. I'll check out the links you had sent and report back.

lib/agent/sql_query_agent/sql_query_agent_answer_prompt.json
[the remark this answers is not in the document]
bborn: +1 for not JSON.

lib/langchain/tool/database.rb
```diff
@@ -0,0 +1,43 @@
...
+  class Database < Base
+    description <<~DESC
+      Useful for getting the result of a database query.
+
+      The input to this tool should be valid SQL.
+    DESC
+
+    # Establish a database connection
+    # example: 'postgres://postgres:postgres@localhost:5432/db_name'
+    # Empty string creates memory database using sqlite3
```
andreibondarev, line 11: Sorry -- why would we create an empty sqlite DB?

lib/langchain/tool/database.rb
```diff
@@ -0,0 +1,43 @@
...
+        Sequel.sqlite
+      else
+        Sequel.connect(db_connection_string)
+      end
+      @db.extension :schema_dumper
+    end
+
+    def schema
+      Langchain.logger.info("Database: Getting schema")
+      @db.dump_schema_migration(same_db: true, foreign_keys: false, indexes: false)
```
andreibondarev, line 28: I think we need to tell the LLM about foreign keys because otherwise it wouldn't know how to join tables together.

lib/langchain/tool/database.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    # Evaluates a sql expression
+    # @param input [String] sql expression
+    # @return [String] results
+    def execute(input:)
+      Langchain.logger.info("Database: Using the Database Tool with \"#{input}\"")
```
andreibondarev, line 35: Would you please modify this to match the format of the following: https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/tool/serp_api.rb#L36?

Tools == light blue
Agents == red

andreibondarev:
This is what `@db.schema` returns: 
```
"Sequel.migration do\n" +
"  change do\n" +
"    create_table(:ar_internal_metadata) do\n" +
"      column :key, \"character varying\", :null=>false\n" +
"      column :value, \"character varying\"\n" +
"      column :created_at, \"timestamp(6) without time zone\", :null=>false\n" +
"      column :updated_at, \"timestamp(6) without time zone\", :null=>false\n" +
"      \n" +
"      primary_key [:key]\n" +
"    end\n" +
"    \n" +
"    create_table(:users) do\n" +
         ...
"    end\n" +
"    \n" +
"    create_table(:schema_migrations) do\n" +
"      column :version, \"character varying\", :null=>false\n" +
"      \n" +
"      primary_key [:version]\n" +
"    end\n" +
"  end\n" +
"end\n" +
```
Probably not the best format to be consumed by the LLM.

rickychilcott:
I can't find documentation anywhere handy, but I watched a Langchain.py talk a month or so ago and based on the conversation they thought the best thing for the LLM would be to provide the schema for the tables in SQL format -- similar to rails `structure.sql` that can be generated via `rails db:schema:dump`.

Ideally, it also includes a few "representative" rows of data -- which is easier said than done. Not sure how to accomplish it easily, but wanted to share.

andreibondarev:
@rickychilcott Yeah, it's in [this paper](https://arxiv.org/pdf/2204.00498.pdf) @bborn linked to.
```sql
CREATE TABLE Highschooler(
ID int primary key,
name text,
grade int)
/*
3 example rows:
SELECT * FROM Highschooler LIMIT 3;
ID name grade
1510 Jordan 9
1689 Gabriel 9
1381 Tiffany 9
*/
CREATE TABLE Friend(
student_id int,
friend_id int,
primary key (student_id,friend_id),
foreign key(student_id) references Highschooler(ID),
foreign key (friend_id) references Highschooler(ID)
)
/*
3 example rows:
SELECT * FROM Friend LIMIT 3;
student_id friend_id
1510 1381
1510 1689
1689 1709
*/
```

My other take-away from that paper was that OpenAI's `davinci-codex` performs better than other models. (I don't have access to that model yet personally).

lib/langchain/tool/database.rb
mattlindsey: My thinking was it creates a usable Database tool when given a blank connection string (which I thought opens up possibities) rather than raising an error, but we can try something else if you want.

lib/langchain/tool/database.rb
mattlindsey: Sounds good.

andreibondarev:
This is a good start (for PostgreSQL):
image

But I guess we'd need to pass it ALL of the tables.

lib/langchain/tool/database.rb
andreibondarev: How difficult would it be to change this to match the schema format that more resembles: https://github.com/andreibondarev/langchainrb/pull/91#issuecomment-1574592840

```ruby
tables = @db.tables
@db["SELECT table_name, column_name, data_type FROM information_schema.columns where table_name in (#{tables})"] # Needs to be reformatted to match ('users', 'user_preferences') etc.
```

mattlindsey:
Can we add these suggestions to a new Issue to be tackled after this PR?

lib/langchain/tool/database.rb
andreibondarev: Was the idea that you can also tell the Agent to "Create me a table for such and such"?

Let's get rid of this use-case, if that's okay, for now. I think it's out of scope of this PR. But it's an interesting concept to keep exploring.

mattlindsey: That was one possibility.  I was also trying to avoid Sequel errors being raised to the calling program/method for now, but that's probably unavoidable.  I'll take out sqlite.

lib/langchain/tool/database.rb
mattlindsey: I think I would have to write that query for every database supported by Sequel, or should I just try it for Postgres for now?

lib/tool/database.rb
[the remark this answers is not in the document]
andreibondarev: > When the number of tables is too big, this won't work (context window), so they dump the table schema into a vector index and query that for the relevant context:

Fascinating... but also complicated. I'm not too sure if we should support it? I guess we could introduce a `persistence:` or `vectorsearch:` object when an `Agent` is initialized that an agent can then use.

langchain.gemspec
```diff
@@ -56,6 +56,8 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "qdrant-ruby", "~> 0.9.0"
   spec.add_development_dependency "ruby-openai", "~> 4.0.0"
   spec.add_development_dependency "safe_ruby", "~> 1.0.4"
+  spec.add_development_dependency "sequel", "~> 5.68.0"
+  spec.add_development_dependency "sqlite3", "~> 1.6.3"
```
andreibondarev, line 60: I think we can get rid of the `sqlite3` gem, right?

lib/langchain/tool/database.rb
mattlindsey: Fixed.

langchain.gemspec
mattlindsey: Fixed.

lib/langchain/tool/database.rb
```diff
@@ -0,0 +1,41 @@
...
+    DESC
+
+    # Establish a database connection
+    # example: 'postgres://user:password@localhost:5432/db_name'
+    def initialize(db_connection_string = "")
+      depends_on "sequel"
+      require "sequel"
+      require "sequel/extensions/schema_dumper"
+
+      @db = if db_connection_string.empty?
```
andreibondarev, line 16: I think this Database Tool should just require the `db_connection_string` and if it's empty or nil or not passed in -- an error should be thrown.

lib/langchain/tool/database.rb
andreibondarev: @mattlindsey If it's okay -- can we just do it for Postgres for now? Other DB vendors in separate issues.

lib/langchain/tool/database.rb
mattlindsey: I'll try, but Sequel throws an akward error.  Do I catch this and return a custom error?

```
  1) Langchain::Agent::SQLQueryAgent#prompt_template returns a prompt template instance
     Failure/Error: @db = Sequel.connect(db_connection_string)

     NoMethodError:
       undefined method `to_sym' for nil:NilClass
     # /Users/mattlindsey/.rvm/gems/ruby-3.2.1/gems/sequel-5.68.0/lib/sequel/database/connecting.rb:17:in `adapter_class'
     # /Users/mattlindsey/.rvm/gems/ruby-3.2.1/gems/sequel-5.68.0/lib/sequel/database/connecting.rb:36:in `connect'
     # /Users/mattlindsey/.rvm/gems/ruby-3.2.1/gems/sequel-5.68.0/lib/sequel/core.rb:124:in `connect'
     # ./lib/langchain/tool/database.rb:16:in `initialize'
     # ./lib/langchain/agent/sql_query_agent/sql_query_agent.rb:17:in `new'
     # ./lib/langchain/agent/sql_query_agent/sql_query_agent.rb:17:in `initialize'
     # ./spec/langchain/agent/sql_query_agent/sql_query_agent_spec.rb:5:in `new'
     # ./spec/langchain/agent/sql_query_agent/sql_query_agent_spec.rb:5:in `block (2 levels) in <top (required)>'
     # ./spec/langchain/agent/sql_query_agent/sql_query_agent_spec.rb:79:in `block (3 levels) in <top (required)>'

Comment (andreibondarev) [reply lib/langchain/tool/database.rb:16]: In the meantime you can just:
```ruby
raise StandardError, "db_connection_string parameter cannot be blank" if db_connection_string.empty?
```

mattlindsey: Done.  Basic functionality is working.  Anything else for this PR?

lib/langchain/tool/database.rb
```diff
@@ -0,0 +1,37 @@
...
+      @db.dump_schema_migration(same_db: true, indexes: false) unless @db.adapter_scheme == :mock
+    end
+
+    # Evaluates a sql expression and outputs labeled values
+    # @param input [String] sql expression
+    # @return [String] results
+    def execute(input:)
+      Langchain.logger.info("[#{self.class.name}]".light_blue + ": Executing \"#{input}\"")
+      @db[input].map do |row|
+        row.map { |key, value| key.to_s + ": " + value.to_s }.each { |col| col.to_s }.join(", ")
```
andreibondarev, line 33: Would it be better to just return `@db[input].to_a` here?
```ruby
[1] pry(#<Langchain::Tool::Database>)> @db[input].to_a
=> [{:id=>8, ... }]
```

mattlindsey: That usually works, but the ai seemed to work better yesterday with multiple row return sets, so I'm thinking a table format with a header might be best.
And some queries seem to be getting off and we get a messy stack trace when the query fails. Maybe a simple message is better?  

```
[LangChain.rb] [Langchain::Tool::Database]: Executing "
SELECT id, name, email, instance_role_id, image_"
/Users/mattlindsey/.rvm/gems/ruby-3.2.1/gems/sequel-5.68.0/lib/sequel/adapters/postgres.rb:171:in `exec': PG::UndefinedColumn: ERROR:  column "id" does not exist (Sequel::DatabaseError)
LINE 2: SELECT id, name, email, instance_role_id, image_
               ^

	from /Users/mattlindsey/.rvm/gems/ruby-3.2.1/gems/sequel-5.68.0/lib/sequel/adapters/postgres.rb:171:in `block in execute_query'
	from /Users/mattlindsey/.rvm/gems/ruby-3.2.1/gems/sequel-5.68.0/lib/sequel/database/logging.rb:38:in `log_connection_yield'
	from /Users/mattlindsey/.rvm/gems/ruby-3.2.1/gems/sequel-5.68.0/lib/sequel/adapters/postgres.rb:171:in `execute_query'
	...

Comment (mattlindsey) [reply lib/langchain/tool/database.rb]: I added more graceful handling.

**Comment by mattlindsey** [comment]
From the conversation seems a good next Issue might be to use Sequel 'reflection' methods to replace the current messy schema dump with 'standard SQL' table definitions and foreign keys (to which sample data can added in a subsequent issues, and the output can be limited to certain tables):
https://sequel.jeremyevans.net/rdoc/files/doc/reflection_rdoc.html

Like this (as suggested by @bborn):
https://github.com/jerryjliu/llama_index/blob/b4618a2a24cd11b5c5949ab97389d62ac34ea336/llama_index/indices/struct_store/container_builder.py#L76
https://github.com/jerryjliu/llama_index/blob/c2f24363b8c6cd74f17647548187821ce9ea4ddf/llama_index/langchain_helpers/sql_wrapper.py#L44

Then later Issues can add a vector store and consider the db vendor by changing the {dialect} in the prompt and including vendor specific schema definition, which would allow the ai to generate better queries I guess.

File: lib/langchain/agent/sql_query_agent/sql_query_agent.rb [hunk 9]
```diff
@@ -1,19 +1,16 @@
 # frozen_string_literal: true
 
 module Langchain::Agent
-  class SQLQueryAgent
+  class SQLQueryAgent < Base
+    attr_reader :llm
+
     # Initializes the Agent
     #
     # @param llm [Symbol] The LLM to use
```
zewelor, line 9: Here [Object] ? And delete next line ?

mattlindsey: Thanks. Fixed.

lib/agent/sql_query_agent/sql_query_agent_answer_prompt.json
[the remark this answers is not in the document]
andreibondarev: https://github.com/andreibondarev/langchainrb/issues/110

lib/tool/database.rb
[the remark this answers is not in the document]
andreibondarev: @bborn ^^ For the cases when the DB schema is too large, unlike LlamaIndex writing to a DB index, I'm actually thinking of implementing it in memory; been looking at Faiss.

PR #98, opened 2023-06-02
Add support for passing additional parameters to OpenAI client constructor

zewelor:
feat(openai.rb): add llm_options in OpenAI client initialization for …

docs(README.md): update documentation for new llm_options parameter

test(openai_spec.rb): add tests for new llm_options parameter in OpenAI client initialization

### DIFF
--- a/lib/langchain/llm/openai.rb
+++ b/lib/langchain/llm/openai.rb
@@ -10,12 +10,12 @@ module Langchain::LLM
       dimension: 1536
     }.freeze
 
-    def initialize(api_key:)
+    def initialize(api_key:, llm_options: {})
       depends_on "ruby-openai"
       require "openai"
 
       # TODO: Add support to pass `organization_id:`
-      @client = ::OpenAI::Client.new(access_token: api_key)
+      @client = ::OpenAI::Client.new(access_token: api_key, **llm_options)
     end
 
     #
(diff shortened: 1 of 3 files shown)


lib/langchain/llm/openai.rb
```diff
@@ -10,12 +10,12 @@ class OpenAI < Base
       dimension: 1536
     }.freeze
 
-    def initialize(api_key:)
+    def initialize(api_key:, llm_options: {})
```
andreibondarev, line 13: Since this is on the `LLM::` class itself, should we just call it `options: {}`?

zewelor: Maybe, but we also have this default hash ( which are also some kind of options ). https://github.com/andreibondarev/langchainrb/pull/98/files#diff-e9b54ad7b6ef063bedb31fba426cea87e7b816d4eeaa65a4ffb07929e6e58646R5

My idea was with llm_options it's more specific this are the options for llm (client?). Not some options for class defined in lib/langchain/llm/openai.rb, but llm client.

but then llm_client_options is kinda long :D. If you think just options will be better name I will make changes.

andreibondarev: Got it, let's keep `llm_options` for now.

PR #101, opened 2023-06-02
Move depends_on out of global namespace

technicalpickles:
Currently, this is in the global namespace, both in terms of file and where the method is defined. In general, it's better to namespace whenever possible.

In the case of a method, we can put it into a module, and then extend it wherever it's needed. That ends up being all the base classes, which makes sense.


andreibondarev (approved):
@technicalpickles 👍🏻

lib/langchain/vectorsearch/base.rb
```diff
@@ -4,6 +4,7 @@
 
 module Langchain::Vectorsearch
   class Base
+    include Langchain::DependencyHelper
```
andreibondarev, line 7: In my other projects I (overkill and) include tests like the following:
```ruby
  describe 'included modules' do
     it { expect(described_class.ancestors.include?(Langchain::DependencyHelper)).to eq(true) }
  end
```

This would go into `vectorsearch/base_spec.rb`

andreibondarev: @technicalpickles It's arguable whether there's any benefit to having a spec like that so I'll leave it up to you.

technicalpickles: Hmm... I'm trying to think through how much that adds 🤔  If you don't include it, then you will end up with no `depends_on` method when you need it, like I saw in the first commit of this. Is that enough for most cases?

I could also imagine checking `expect(subject).to respond_to(:depends_on)`, so you get a single method that tells you what is wrong.

Another option could be making a method like `load_dependencies`, which ends up calling. That way, you end up with a test that makes says class is able to load its dependencies 🤔 🤔

andreibondarev: True, `described_class.new()` would just bomb. I'm okay leaving it out.

andreibondarev:
@technicalpickles Lmk if you'd like me to merge?

technicalpickles:
@andreibondarev :shipit:

PR #102, opened 2023-06-02
Split standardrb to its own CI step

technicalpickles:
Currently, CI runs standardrb and then tests, so if standardb fails you don't get any feedback from the test.

It also runs standardb on all 3 versions of ruby. In practice, you really only need to run it on one version.

This splits `linters` out from `tests`. I tried to mirror the naming, but I'm open to explicitly calling it standardrb.

Lastly, I'm trying to turn on `--format github` which HOPEFULLY will start annotating the PR with the failure. We shouldn't have to rely on a maintainer to point out that standardrb --fix is needed :sweat_smile:

Oh, and since we are hear, also turn on color output since GitHub Actions supports it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@master
-
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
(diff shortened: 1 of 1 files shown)


andreibondarev:
Amazing... I didn't even think about this.

technicalpickles:
Here's what it looks like on the Files changed panel:

CleanShot 2023-06-02 at 16 45 51

andreibondarev (approved):
@technicalpickles Super cool!

PR #106, opened 2023-06-03
Add interface for other components to build an instance of an LLM

technicalpickles:
Currently, we have a few places that instantiate an LLM class (vectorsearch and agent).

The current implementation ends up looking at
a constant to find a mapping from a name to a class name, and then uses const_get on Langchain::LLM::Base to get instantiate it. They do use a method from Base to make sure it's valid first. But even with that, the component using an LLM ends up pretty coupled to the LLM data structure for tracking what it supports.

Since we are using a name, and building an instance, it made me think factory pattern. And thinking factory pattern, reminded me of factory_bot. That inspired this interface:

    Langchain::LLM.build(:openai, api_key: "123")

I almost left it there, but something that has been bugging me is how to allow third party LLMs that aren't part of the repo. The current way does requires it to be in that hash of names to class names. I got to thinking, maybe we can support passing in an LLM object directly? In that case, we don't need to do anything to look it up.

This actually lines up with how langchain.py is simplemented:

    llm = OpenAI(temperature=0.9)
    prompt = PromptTemplate(
        input_variables=["product"],
        template="What is a good name for a company that makes {product}?",
    )
    chain = LLMChain(llm=llm, prompt=prompt)

I ended up adding these to the LLM module directly, and moved some logic from LLM::Base, which left that class solely responsible for a base class, instead of also being responsible for validating and knowing what LLMs are supported.


andreibondarev:
@zewelor had literally brought this up today. @zewelor did you already build something out?

@technicalpickles Btw, a lot of us are in the [Ruby AI Builders](https://discord.gg/SBmjAnKT) Discord.

lib/langchain/agent/chain_of_thought_agent/chain_of_thought_agent.rb
```diff
@@ -6,19 +6,18 @@ class ChainOfThoughtAgent < Base
...
     # Initializes the Agent
     #
-    # @param llm [Symbol] The LLM to use
-    # @param llm_api_key [String] The API key for the LLM
-    # @param tools [Array] The tools to use
+    # @param llm [Symbol] Name of the LLM to use
+    # @param llm_api_key [String] API key for the LLM
+    # @param tools [Array] List of tools to use, as Strings
     # @return [ChainOfThoughtAgent] The Agent::ChainOfThoughtAgent instance
     def initialize(llm:, llm_api_key:, tools: [])
```
andreibondarev, line 13: I think this needs to be modified to `llm_api_key: nil` because if you're passing an `llm: llm_object`, there's no need to pass in the api key.

andreibondarev:
@technicalpickles I do like being able to pass the object around instead of re-instantiating a new LLM object every single time.

With the factory pattern... if we implement it here, do you feel like we ought to extend it to everything else in the project?

zewelor:
Hi

Yes same idea I've wrote yesterday on discord. I think its step in right direction. Here is my finished implementation: https://github.com/andreibondarev/langchainrb/pull/109

I've changed it, after looking here to extract build logic to separate module, just like done here. I guess main difference is whether to still allow to pass llm, and llm_key down to the classes, and then build_or_resuse like here. From my point of view if class has less arguments its better because its simpler. Its also somehow undefined behavior now, when someone passes llm: as build object and also llm_key: . What should happen then ?
Class doesn't need, to know, what to do with different kind of arguments. Logic is extracted but still in constructor, each class needs to know to call build_or_reuse etc. We still provide helper class to quick build object, like before, but interfaces in rest of the code are simpler when sticking with llm_client. Thats my reasoning, happy to discuss of course.

technicalpickles:
> With the factory pattern... if we implement it here, do you feel like we ought to extend it to everything else in the project?

@andreibondarev Yeah, that would make sense to me. And in general, support passing around objects, rather than specifying names as symbols. LLM is just where I first saw it in practice.

> Yes same idea I've wrote yesterday on discord.

@zewelor Can I ask which Discord? I don't see any mentioned in the repo or docs or whatever.

> I've changed my code, after looking here, to extract build logic to separate module, just like done here. Cleaner than my initial approach !

Nice. Can you make sure to credit this PR if that's the case?

> I guess main difference is whether to still allow to pass llm, and llm_key down to the classes, and then build_or_resuse like here. From my point of view if class has less arguments its better because its simpler. Its also somehow undefined behavior now, when someone passes llm: as build object and also llm_key: . What should happen then ?

Good points. I think in practice, people would probably pass around the same key. I think passing in an llm_key when there's an LLM object would be an error. That logic can live in build_or_reuse.

> Class doesn't need, to know, what to do with different kind of arguments. Logic is extracted but still in constructor, each class needs to know to call build_or_reuse etc. We still provide helper class to quick build object, like before, but interfaces in rest of the code are simpler when sticking with llm_client. Thats my reasoning, happy to discuss of course.

Between classes needing to know to always call build_or_reuse, and that the constructors take extra arguments for LLM keys, is a sign that we should consolidate on one key, the LLM object. To draw comparison to langchain.py again, that is how it is implemented.

I would recommend using `llm` as the key, rather than  `llm_client` though. I think the `client` doesn't add much, because for most practical purposes, a developer can think of the client as the LLM, at this level of abstraction.

andreibondarev:
@technicalpickles [Ruby AI Builders](https://discord.gg/SBmjAnKT)

zewelor:
> > I've changed my code, after looking here, to extract build logic to separate module, just like done here. Cleaner than my initial approach !
> 
> Nice. Can you make sure to credit this PR if that's the case?
> 

After request from reviewer and @andreibondarev I've removed build logic completely in my PR now. Sorry for not linking it here right away.

technicalpickles:
Closing in favor of other work that landed. I'll review and see if there's anything else worth pulling out here still.

PR #109, opened 2023-06-03
Allow to pass llm_client object

zewelor:
As discussed in discord, added ability to build llm_client externally and to pass it, instead of creating llm_client object in each class. This way we can open more people to construct their own llm_clients and use them. Also its possible to change some default in classes, like used model in OpenAI etc.


lib/langchain/llm.rb
```diff
@@ -0,0 +1,35 @@
...
+      openai: "OpenAI",
+      replicate: "Replicate"
+    }.freeze
+
+    def build(llm, api_key)
+      unless supported?(llm)
+        raise ArgumentError, "LLM must be one of #{self::BUILTIN.keys}"
+      end
+
+      const_get(self::BUILTIN.fetch(llm)).new(api_key: api_key)
```
alchaplinsky, line 28: It is a good move to allow passing LLM client instance to all these various services instead of a symbol 👍  
However, I don't see much value in having and using this `build` method and restricting clients only to the list implemented in `langchainrb` codebase. I believe we should allow passing any LLM client instance, even if it is a custom implementation of a user of this library. So instead of:

```ruby
weaviate = Langchain::Vectorsearch::Weaviate.new(
  url: ENV["WEAVIATE_URL"],
  index_name: "",
  llm_client: Langchain::LLM.build(:openai, ENV["OPENAI_API_KEY"])
)
```

user could do:

```ruby
llm_client = Langchain::LLM.OpenAI.new(api_key: ENV["OPENAI_API_KEY"])
weaviate = Langchain::Vectorsearch::Weaviate.new(
  url: ENV["WEAVIATE_URL"],
  index_name: "",
  llm_client: llm_client
)
```
Some pros of this approach:
- Restricting LLM clients to a list that is implemented does not make sense, as developers might implement custom clients to their own LLMs or any LLM service that is not yet supported by langchainrb. Should we restrict them from using a library with their own LLM clients? I think no. Eventually, they might want to contribute and add a new client to the library's codebase.
- Adding a new LLM client would be as easy as adding a class and spec for it. No need to update `lib/langchain/llm.rb` file and make sure that `build` method works properly with `:new_client` symbol.

alchaplinsky: Let's get rid of passing `:openai` as a symbol and mapping `:openai` to `LLM::OpenAI`. It is better to manipulate with instances.
cc @andreibondarev

zewelor: My idea is to don't restrict, you can still create any class you want and use it. This builder keep part of old logic, but nothing prevents creating own class and using it as llm_client. I assumed we might want to keep some syntax sugar, but maybe not ?

```
llm_client = Some::Custom::OpenAi.OpenAI.new(custom_args) ( at least if it implements required methods )
weaviate = Langchain::Vectorsearch::Weaviate.new(
  url: ENV["WEAVIATE_URL"],
  index_name: "",
  llm_client: llm_client
)
```

andreibondarev:
@zewelor @technicalpickles Okay -- can I suggest something to make these changes manageable? 

Can we make this PR simply converting the previous `llm:, llm_api_key:` pattern to `llm: llm_object` and revert the builder pattern?

zewelor:
I've removed buil method. Please take a look now if its ok.

andreibondarev:
@zewelor Sorry -- one last thing: Do you think it's better if the name of the param is just `llm:` instead of `llm_client:`?

lib/langchain/llm.rb
alchaplinsky: Oh, got it. I thought it was only possible to instantiate an LLM via `build` method.

zewelor:
@andreibondarev Yes, changed llm_client to llm everywhere. I hope now its ok.

alchaplinsky:
LGTM, nice job @zewelor!

PR #111, opened 2023-06-03
Allow passing a model as an option to OpenAI client

alchaplinsky:
## Motivation
Update `embed`, `complete`, and `chat` methods on the OpenAI client to allow passing `model` as a param in case if a user wants to use a different model from the default one.

## Changes
- Extract parameters composition into a private method. We need to come up with some general algorithm for extending default params with passed params and put it into the `Base` class.
- Update specs to test params that are sent to OpenAI client

### DIFF
--- a/lib/langchain/llm/openai.rb
+++ b/lib/langchain/llm/openai.rb
@@ -23,17 +23,12 @@ module Langchain::LLM
     # @param text [String] The text to generate an embedding for
     # @return [Array] The embedding
     #
-    def embed(text:)
-      model = DEFAULTS[:embeddings_model_name]
+    def embed(text:, **params)
+      parameters = {model: DEFAULTS[:embeddings_model_name], input: text}
 
-      Langchain::Utils::TokenLengthValidator.validate!(text, model)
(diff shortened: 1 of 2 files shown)


lib/langchain/llm/openai.rb
```diff
@@ -23,17 +23,12 @@ def initialize(api_key:, llm_options: {})
     # @param text [String] The text to generate an embedding for
     # @return [Array] The embedding
     #
-    def embed(text:)
-      model = DEFAULTS[:embeddings_model_name]
+    def embed(text:, **params)
+      parameters = {model: DEFAULTS[:embeddings_model_name], input: text}
```
andreibondarev, line 27: So you can't overwrite the model used in this `embed()` method?

alchaplinsky: You can override by passing it in the params. On line `31` I'm doing a merge to these parameters.

andreibondarev: Oops -- didn't see it!

andreibondarev (approved):

Issue #54, opened 2023-05-24
Improve the Langchainrb documentation


andreibondarev:
### Problem
We have outgrown the single-file README. It is time to migrate to a more scalable solution.

Auto-generated documention RDoc-style may not be ideal as it'll read as a dry technical manual.

### Ideas
- Create a dedicated a website (@rickychilcott shared https://www.phlex.fun/ as an example)

andreibondarev:
@ZhenhangTung I know you had mentioned that the documentation was lacking. Would [this](https://github.com/andreibondarev/langchainrb/wiki) be a better approach? Do you have any feedback on how to improve it further?

ZhenhangTung:
@andreibondarev I do think it's better than the previous one. It's more structured and guide uses with concepts and code. I'll provide feedback if I have any in the future. For now, I believe we should prioritize something like user investigations or feature implementations.

technicalpickles:
> Move documentation to Wiki

The main downsides are that you can't do PRs against them, and that they are totally separate from the main repo. That means it's an extra step to update if code changes means docs changes.

> Auto-generated documention RDoc-style may not be ideal as it'll read as a dry technical manual.

There is definitely value in it though. I've had to go to the langchain(.py) docs, and found them hard to read, and not super useful. Example: https://python.langchain.com/en/latest/reference/modules/vectorstores.html

Rails itself is a good example of thorough API reference (https://api.rubyonrails.org) , while there is still room for more general guides (https://guides.rubyonrails.org/)

andreibondarev:
> https://python.langchain.com/en/latest/reference/modules/vectorstores.html
Yeah, that's exactly the technical manual look and feel that I was referring to.

I really like how Rails is documented. And not just the docs but the code itself!

And you're right, I think we'll need to create a `/docs` directory with a bunch markdown files so they can be modified along with actual code changes. And then auto-deploy the docs to Github Pages or something.

technicalpickles:
A few changes landed since last update: https://github.com/andreibondarev/langchainrb/pull/112  https://github.com/andreibondarev/langchainrb/pull/152

https://rubydoc.info/gems/langchainrb now contains nearly all of the contents of the wiki, but located on the relevant class's rdoc.

PR #112, opened 2023-06-03
Add YARD for building documentation

technicalpickles:
Some work around https://github.com/andreibondarev/langchainrb/issues/54 for when we want to publish an API reference. I've confirmed this working locally, run with `bundle exec rake yard`, and then `open doc/index.html`

One thing I wasn't able to figure out was getting the emoji short codes in the table working. Regular emoji characters work though.

I did initially try to access on rubydoc.info, ie https://rubydoc.info/gems/langchainrb ... but something is definitely not right. Not sure if it's related to the repo, or something with that.

I can try again after this lands and is released to see if it helps. I can file an issue on https://github.com/docmeta/rubydoc.info if it still is around

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,5 +60,5 @@ jobs:
           ruby-version: 3.2
           bundler: default
           bundler-cache: true
-      - name: StandardRb check
+      - name: Build docs
         run: bundle exec rake yard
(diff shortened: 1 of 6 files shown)


andreibondarev (approved):
@technicalpickles Another amazing piece of work from your sir! 👏🏻

Issue #62, opened 2023-05-25
[Proposal] Change method add_data's parameter


ZhenhangTung:
Under module `Vectorsearch::Base`, there is a method `add_data` accepting `path: nil, paths: nil` params. 
From my point of view, we could remove `path` and only accept `paths` or use the [splat operator](https://medium.com/@sologoubalex/parameter-with-splat-operator-in-ruby-part-1-2-a1c2176215a5).

andreibondarev:
@ZhenhangTung I'll let the @rickychilcott chime in as he built the original implementation. I don't have a preference here.

rickychilcott:
I'd be supportive of it becoming:

```ruby
def add_data(paths:)
  texts = 
    Array(paths)
      .map { |path| Langchain::Loader.new(path)&.load&.value }
      .compact

  add_texts(texts: texts)
end
```

I'd like to keep the named parameter:

a) to be consistent with the API elsewhere in the class
b) to provide an opportunity to support other arguments

ZhenhangTung:
@rickychilcott Great!

andreibondarev:
@ZhenhangTung Would you have any interest and availability to open up a PR for this?

ZhenhangTung:
@andreibondare That would be my honor.

PR #118, opened 2023-06-04
Improve add_data's implementation

ZhenhangTung:
This PR is the implementation of https://github.com/andreibondarev/langchainrb/issues/62.


lib/langchain/vectorsearch/base.rb
```diff
@@ -63,14 +63,13 @@ def generate_prompt(question:, context:)
       prompt_template.format(question: question)
     end
 
-    def add_data(path: nil, paths: nil)
-      raise ArgumentError, "Either path or paths must be provided" if path.nil? && paths.nil?
-      raise ArgumentError, "Either path or paths must be provided, not both" if !path.nil? && !paths.nil?
-
-      texts = Array(path || paths)
-        .flatten
```
andreibondarev, line 71: I think we still need `.flatten`, no? It covers the cases when someone passes in a 2d array:

```ruby
irb(main):002:0> Array([[]]).flatten
=> []
irb(main):003:0> Array([]).flatten
=> []
```

ZhenhangTung: I believe such cases rarely occur if we don't tell users to do so, but yeah we could keep `.flatten` to handle them. Will update the code tonight 😄 .

ZhenhangTung: @andreibondarev The code was updated. Please review it when you have time, thanks!

andreibondarev (approved):

andreibondarev:
@ZhenhangTung Thank you! 👏🏻

PR #121, opened 2023-06-04
Support loading prompts from yaml

zewelor:
Added support to load prompt from yaml file


spec/fixtures/prompt/prompt_template.yaml
```diff
@@ -0,0 +1,5 @@
+_type: prompt
+input_variables:
+  - adjective
+  - content
+template: Tell me a {adjective} joke about {content}.
```
andreibondarev, line 5: Does yaml support multi-line strings?

andreibondarev: That was my complaint with JSON -- that you can't do multiline strings and a lengthy template would just be a single line with tons of line break characters.

zewelor:
Yes in YAML its quite easy to do multiline string. Changes spec fixture to show an example. Here is some site to take a look at different options how multiline strings can be used in yaml: https://yaml-multiline.info/

alchaplinsky:
Yeah, I think `yaml` is better format for prompts than JSON. Prompts will be more readable.

andreibondarev:
@zewelor This is great, thank you!

PR #122, opened 2023-06-04
Reduce max_tokens in SQLQueryAgent first step

mattlindsey:
Prompt + completion was apparently sometimes exceeding the limit (see below).  200 should be good to return even a large sql_string.
Possible enhancement would be to calculate it:  model_limit - prompt_token_size.

```
max_tokens
integer
Optional
Defaults to 16
The maximum number of [tokens](https://platform.openai.com/tokenizer) to generate in the completion.

The token count of your prompt plus max_tokens cannot exceed the model's context length. [Example Python code](https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb) for counting tokens.
```


andreibondarev:
@mattlindsey Maybe instead of merging this in, we'd tackle this instead? https://github.com/andreibondarev/langchainrb/issues/123

PR #124, opened 2023-06-05
add ability to pass uuid

jackpaulcollins:


jackpaulcollins:
Oops sorry meant to open this on my fork

andreibondarev:
You did open it on your fork.

lib/langchain/vectorsearch/weaviate.rb
```diff
@@ -32,7 +32,7 @@ def initialize(url:, api_key:, index_name:, llm:)
     # Add a list of texts to the index
     # @param texts [Array] The list of texts to add
     # @return [Hash] The response from the server
-    def add_texts(texts:)
+    def add_texts(texts:, uuid:,)
```
andreibondarev, line 35: I think this is definitely needed, but I think it's worth double checking other vector search classes as well to make sure that we've got the same interface for all of them. And maybe "ids:" would be more universal?

jackpaulcollins: Oh weird, this made me nervous. I'm not 100% confident I understand how forking works.

Screenshot 2023-06-04 at 9 13 39 PM

>but I think it's worth double checking other vector search classes as well to make sure that we've got the same interface for all of them.

Agreed. My goal at the moment is to learn how the weaviate api works. It isn't immediately clear to me what happens when a vector blob is uploaded, and how _that particular _blob__ can be queried in the future.

My use case is a document upload service in which documents are uploaded via a rails app. When a doc's uploaded the rails app will enqueue a worker to upload the doc text to weaviate and save it (`#<Document id, title, weaviate_uuid:>`), allowing a user to visit documents at their leisure and perform LLM queries against them. This is why the uuid name made sense to me ([that's also what they call it](https://weaviate.io/developers/academy/units/schema_and_imports/import#optional-parameters))

I'm assuming most other clients will be similar? 
 
Let me know if forking the repos (LchainRB and weaviate-ruby) to experiment isn't an ideal approach, I don't want ya'll to be notified etc with PRs that aren't meaningful to you while I experiment/learn 🙃

andreibondarev: > I'm assuming most other clients will be similar?

Some other vector search DBs call that field ID not UUID. I just meant that "ID" felt to be more universal than "UUID".

> Let me know if forking the repos (LchainRB and weaviate-ruby) to experiment isn't an ideal approach, I don't want ya'll to be notified etc with PRs that aren't meaningful to you while I experiment/learn 🙃

Yeah, go for it! Actually no one is getting a notification when you're playing around with your own fork unless you open a PR against origin.

PR #125, opened 2023-06-05
Support rubocop tooling with standard rules

zewelor:
This way its possible to use rubocop cli, for example rubocop -A, and gets all of the standardrb rules applied. In my case some vscode extenstions doesnt know about standard, but works well with rubocop. Using this all just works and rubocop see the same rules.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,11 @@
+AllCops:
+  SuggestExtensions: false
+
+require:
+  - standard
+  - rubocop-performance
+
+inherit_gem:
+  standard: config/base.yml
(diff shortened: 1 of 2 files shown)


rickychilcott:
This seems reasonable. Should we add `rubocop` as an explicit dependency? standardrb will never get rid of that dependency, but since we're using it directly now, it seems wise to add it.

zewelor:
Added it with some comment to avoid confusion

andreibondarev (approved):

andreibondarev:
@zewelor 👍🏻

PR #126, opened 2023-06-05
Add lefthook for auto lint before commit etc

zewelor:
Adds lefthook for pre-commit and other git hooks management. Now on pre-commit it will auto lint code ( so faster feedback about not formatted code etc ). 

It can be enabled as option.

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -1,0 +1,4 @@
+pre-commit:
+  commands:
+    lint:
+      run: standardrb --fix
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,11 @@
+AllCops:
+  SuggestExtensions: false
+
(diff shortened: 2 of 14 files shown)


zewelor:
Good catch, removed it from gemspec and added gem install to readme

lefthook.yml
```diff
@@ -0,0 +1,4 @@
+pre-commit:
+  commands:
+    lint:
+      run: standardrb --fix
```
technicalpickles, line 4: You may need also need `stage_fixed: https://github.com/evilmartians/lefthook/blob/master/docs/configuration.md#stage_fixed

technicalpickles (commented):
Thanks for this! It is something I was thinking about because of the times I had CI fail with standardrb fixes 🙃

PR #130, opened 2023-06-05
Change Agent interface for Tools

### LINKED ISSUE #113: Modify Tool usage to pass tool instances instead of tool names as strings.
Currently when an Agent is initialized to use Tools we pass them as strings that are then [matched to existing classes](https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/tool/base.rb#L14-L18). See the following current usage:
```ruby
agent = Langchain::Agent::ChainOfThoughtAgent.new(llm: :openai, llm_api_key: ENV["OPENAI_API_KEY"], tools: ['search', 'calculator'])

agent.tools
# => ["search", "calculator"]

agent.tools = ["wikipedia"]
```

This approach is not flexible because:
* We're unable to configure existing Tools
* We're unable to pass new Tools to Agents

Let's change the Agent interface to accept Tool instances like so:
```ruby
calculator_tool = Langchain::Tool::Calculator.new()
sql_db_tool = Langchain::Tool::Database.new(db_connection_string: "postgres://user:password@localhost:5432/db_name") # Coming in this PR: https://github.com/andreibondarev/langchainrb/pull/91/files#diff-9a2d0c4b8a1176be3d78866742f2ba4c2da2452cb05959f433c1204bb8211ebd

agent = Langchain::Agent::ChainOfThoughtAgent.new(
  llm: :openai,
  llm_api_key: ENV["OPENAI_API_KEY"],
  tools: [calculator_tool, sql_db_tool]
)
```

Note: Modify the SerpApi tool to accept the `api_key:` in the initialize method instead of seeking out the ENV var [here](https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/tool/serp_api.rb#L52):
```ruby
module Langchain::Tool
  class SerpApi < Base
  
    attr_reader :api_key

    def initialize(api_key:)
      @api_key = api_key

    def execute_search(input:)
      GoogleSearch.new(
        serp_api_key: api_key
```

mattlindsey:
Change the Agent interface to accept Tool instances.
Fixes #113

WIP.  I need to update the README


lib/langchain/tool/base.rb
```diff
@@ -7,15 +7,15 @@ class Base
...
 
     TOOLS = {
-      "calculator" => "Langchain::Tool::Calculator",
-      "search" => "Langchain::Tool::SerpApi",
-      "wikipedia" => "Langchain::Tool::Wikipedia",
-      "database" => "Langchain::Tool::Database"
+      "Langchain::Tool::Calculator" => "calculator",
+      "Langchain::Tool::SerpApi" => "search",
+      "Langchain::Tool::Wikipedia" => "wikipedia",
+      "Langchain::Tool::Database" => "database"
```
andreibondarev, line 18: What if instead of this hash map, we just had a `name` or `tool_name` method on the Tool class itself? For example you can call the `DESCRIPTION` constant on each tool:
```ruby
Langchain::Tool::Calculator::DESCRIPTION
```

lib/langchain/tool/base.rb
```diff
@@ -38,15 +38,15 @@ def execute(input:)
 
     #
     # Validates the list of strings (tools) are all supported or raises an error
-    # @param tools [Array<String>] list of tools to be used
+    # @param tools [Array<Langchain::Tool>] list of tools to be used
     #
     # @raise [ArgumentError] If any of the tools are not supported
     #
     def self.validate_tools!(tools:)
```
andreibondarev, line 45: I'm not sure we need to validate these tools anymore because, technically, I could just write up my own tool with the `#execute()` method, a description, and name, and pass it in to the Agent.

lib/langchain/agent/chain_of_thought_agent/chain_of_thought_agent.rb
```diff
@@ -81,12 +81,18 @@ def run(question:)
     # @param tools [Array] Tools to use
     # @return [String] Prompt
     def create_prompt(question:, tools:)
+      tool_list = tools.map do |tool|
```
andreibondarev, line 84: If we had the `.name` method this would just become `tool_list = tools.map(&:name)`

lib/langchain/tool/base.rb
mattlindsey: So we want it allowed to run any tool, regardless of being 'registered.'  You think we can take out the validate and and raise_error entirely?

lib/langchain/tool/base.rb
mattlindsey: I thought maybe this was a safety mechanism to avoid passing in any 'unregistered' tool_name.  See below.

lib/langchain/tool/base.rb
andreibondarev: Yes! I think the approach is "we have a few useful tools but if you're missing what you need -- easily build and connect whatever tool you'd like".

andreibondarev:
```
autoload :SerpApi, "langchain/tool/serp_api"
```
^^ This, and for other existing tools, is still needed.

The Agent will not be instantiating any new tool, it just uses the tool instances that were passed in via `tools: [...]`.

lib/langchain/tool/calculator.rb
```diff
@@ -33,6 +33,7 @@ def execute(input:)
     rescue Eqn::ParseError, Eqn::NoVariableValueError
       # Sometimes the input is not a pure math expression, e.g: "12F in Celsius"
       # We can use the google answer box to evaluate this expression
+      # TODO: SERP_API_KEY is now being passed into SerpApi constructor, so I think below line won't work as is
       hash_results = Langchain::Tool::SerpApi.execute_search(input: input)
```
mattlindsey, line 37: @andreibondarev Everything seems to work except this line.  It's failing on the api_key and I don't think the Calculator tool should be calling this directly, but I'm unsure how to fix it.  Might need to pass in the search instance to Calculator#execute?

lib/langchain/agent/chain_of_thought_agent/chain_of_thought_agent.rb
mattlindsey: I added a NAME constant to each tool.

.gitignore
```diff
@@ -6,10 +6,13 @@
...
 /tmp/
+/.vscode/
 
 # rspec failure tracking
 .rspec_status
 
 .env
 .DS_Store
 .idea
+.rubocop.yml
```
andreibondarev, line 17: Why do we need this change ignoring the `.rubocop.yml` file? We definitely want git to keep tracking it.

lib/langchain/agent/chain_of_thought_agent/chain_of_thought_agent.rb
```diff
@@ -55,10 +55,11 @@ def run(question:)
           # Find the input to the action in the "Action Input: [action_input]" format
           action_input = response.match(/Action Input: "?(.*)"?/)&.send(:[], -1)
 
-          # Retrieve the Tool::[ToolName] class and call `execute`` with action_input as the input
-          tool = Langchain::Tool.const_get(Langchain::Tool::Base::TOOLS[action.strip])
-          Langchain.logger.info("[#{self.class.name}]".red + ": Invoking \"#{tool}\" Tool with \"#{action_input}\"")
+          # Retrieve the Tool::[ToolName] class from tools with the name of the action
+          tool = tools.find { |tool| Langchain::Tool.const_get(tool.class.to_s).const_get(:NAME) == action.strip }
```
andreibondarev, line 59: I think you should be able to do:
```ruby
tool = tools.find { |tool| tool.class.const_get(:NAME) == action.strip }
```

lib/langchain/agent/chain_of_thought_agent/chain_of_thought_agent.rb
```diff
@@ -81,12 +82,18 @@ def run(question:)
...
+      end.join(", ")
+
       prompt_template.format(
         date: Date.today.strftime("%B %d, %Y"),
         question: question,
-        tool_names: "[#{tools.join(", ")}]",
+        tool_names: "[#{tool_list}]",
         tools: tools.map do |tool|
-          "#{tool}: #{Langchain::Tool.const_get(Langchain::Tool::Base::TOOLS[tool]).const_get(:DESCRIPTION)}"
+          tool_name = Langchain::Tool.const_get(tool.class.to_s).const_get(:NAME)
```
andreibondarev, line 94: This can just be:
```ruby
tool_name = tool.class.const_get(:NAME)
tool_description = tool.class.const_get(:DESCRIPTION)
```

lib/langchain/tool/calculator.rb
```diff
@@ -10,7 +10,7 @@ class Calculator < Base
     #   gem "google_search_results", "~> 2.0.0"
     # ENV requirements: ENV["SERPAPI_API_KEY"]
     #
-
+    const_set(:NAME, "calculator")
```
andreibondarev, line 13: Please change to:
```ruby
NAME = "calculator"
```

lib/langchain/tool/calculator.rb
andreibondarev: Actually -- don't worry about it for now, I'll probably just get rid of this.

lib/langchain/tool/wikipedia.rb
```diff
@@ -7,7 +7,7 @@ class Wikipedia < Base
     #
     # Gem requirements: gem "wikipedia-client", "~> 1.17.0"
     #
-
+    const_set(:NAME, "wikipedia")
```
andreibondarev, line 10: NAME = "wikipedia"

lib/langchain/tool/serp_api.rb
```diff
@@ -6,9 +6,10 @@ class SerpApi < Base
     # Wrapper around SerpAPI
     #
     # Gem requirements: gem "google_search_results", "~> 2.0.0"
-    # ENV requirements: ENV["SERPAPI_API_KEY"] # https://serpapi.com/manage-api-key)
     #
+    attr_reader :api_key
 
+    const_set(:NAME, "search")
```
andreibondarev, line 12: NAME = "search"

lib/langchain/tool/ruby_code_interpreter.rb
```diff
@@ -7,7 +7,7 @@ class RubyCodeInterpreter < Base
     #
     # Gem requirements: gem "safe_ruby", "~> 1.0.4"
     #
-
+    const_set(:NAME, "ruby-interpreter")
```
andreibondarev, line 10:
```ruby
NAME = "ruby_code_interpreter"
```

Let's keep the old name for now.

lib/langchain/tool/database.rb
```diff
@@ -5,7 +5,7 @@ class Database < Base
     #
     # Gem requirements: gem "sequel", "~> 5.68.0"
     #
-
+    const_set(:NAME, "database")
```
andreibondarev, line 8:
```ruby
NAME = "database"
```

lib/langchain/tool/calculator.rb
mattlindsey: OK.  I made the changes you noted so far, but there's 1 test marked 'xit' now because it will fail until this is fixed.

andreibondarev:
@mattlindsey Great job on this PR! Thank you!

I tweaked a few things... merging now! 🚀

Issue #108, opened 2023-06-03
Debugging this project doesn't allow 'step' or 'continue'


mattlindsey:
When I set a breakpoint with 'binding.pry' most debugging commands work except 'step' and 'continue.'

```
[2] pry(#<Langchain::Tool::Database>)> whereami
Inside #<Langchain::Tool::Database>.
[3] pry(#<Langchain::Tool::Database>)>
[4] pry(#<Langchain::Tool::Database>)> step
NameError: undefined local variable or method `step' for #<Langchain::Tool::Database:0x00000001040e0e20 @DB=#<Sequel::SQLite::Database: {:adapter=>:sqlite}>>
from (pry):1:in `__pry__'
```

PR #134, opened 2023-06-05
Require pry-byebug to allow step, next, continue

alchaplinsky:
Fixes https://github.com/andreibondarev/langchainrb/issues/108

Details are nicely described in the article: https://blog.pawandubey.com/explained-binding-pry-byebug-debugger

TL;DR: It's all more complicated than it should be.


andreibondarev:
Awesome! 

I know @mattlindsey will be very happy!

PR #141, opened 2023-06-06
Add util method to calculate max_tokens

### LINKED ISSUE #123: Calculate max_tokens based on model and prompt token values
Prompt + completion is apparently sometimes exceeding with hardcoded max_token values in the agents.
I think it can be calculated as: model_limit - prompt_token_size.  
Use the Utils class TokenLengthValidator and/or Tiktoken.

(https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb) for counting tokens.](https://platform.openai.com/docs/api-reference/completions/create#completions/create-max_tokens)

mattlindsey:
Fixes #123 

Checking if this is the right approach.  Note that tiktoken will be called twice if this is used.


mattlindsey:
Something cleaner would be to NOT add `attr_reader :model_name` to LLM::Base, but only add code to LLM::Base::OpenAI whereby 'max_tokens: -1' to a method indicates that you want to have that method calculate max_tokens.  Any value other than -1 would pass that directly on to openai.

PR #143, opened 2023-06-07
Improve db schema dump

### LINKED ISSUE #138: Improve db tool schema dump
As a prequel to #129, use Sequel 'reflection' methods to replace the current messy schema dump with clean table definitions and foreign keys (to which sample data can added in a subsequent issues, and the output can be limited to certain tables):
https://sequel.jeremyevans.net/rdoc/files/doc/reflection_rdoc.html

Like this (as suggested by @bborn):
https://github.com/jerryjliu/llama_index/blob/b4618a2a24cd11b5c5949ab97389d62ac34ea336/llama_index/indices/struct_store/container_builder.py#L76
https://github.com/jerryjliu/llama_index/blob/c2f24363b8c6cd74f17647548187821ce9ea4ddf/llama_index/langchain_helpers/sql_wrapper.py#L44

Format (suggested by @rickychilcott from [this paper](https://arxiv.org/pdf/2204.00498.pdf))
``` sql
CREATE TABLE Highschooler(
ID int primary key,
name text,
grade int)
/*
3 example rows:
SELECT * FROM Highschooler LIMIT 3;
ID name grade
1510 Jordan 9
1689 Gabriel 9
1381 Tiffany 9
*/
CREATE TABLE Friend(
student_id int,
friend_id int,
primary key (student_id,friend_id),
foreign key(student_id) references Highschooler(ID),
foreign key (friend_id) references Highschooler(ID)
)
/*
3 example rows:
SELECT * FROM Friend LIMIT 3;
student_id friend_id
1510 1381
1510 1689
1689 1709
*/

### PR DESCRIPTION
**Pull request by mattlindsey**
Improves the database schema dump to easy to digest format.
Fixes #138

### DIFF
--- a/lib/langchain/tool/database.rb
+++ b/lib/langchain/tool/database.rb
@@ -39,8 +39,24 @@ module Langchain::Tool
     # @return [String] schema
     #
     def schema
+      return if db.adapter_scheme == :mock
       Langchain.logger.info("[#{self.class.name}]".light_blue + ": Dumping schema")
-      db.dump_schema_migration(same_db: true, indexes: false) unless db.adapter_scheme == :mock
+
+      schema = ""
+      db.tables.each do |table|
+        schema << "CREATE TABLE #{table}(\n"
+        db.schema(table).each do |column|
+          schema << "#{column[0]} #{column[1][:type]}"
+          schema << " PRIMARY KEY" if column[1][:primary_key] == true
+          schema << "," unless column == db.schema(table).last
+          schema << "\n"
+        end
+        schema << ");\n"
+        db.foreign_key_list(table).each do |fk|
+          schema << "ALTER TABLE #{table} ADD FOREIGN KEY (#{fk[:columns][0]}) REFERENCES #{fk[:table]}(#{fk[:key][0]});\n"
+        end
+      end
+      schema
     end
 
     #

### CODE REVIEW

**Comment by mattlindsey** [comment]
This PR can go in as-is if approved, but adding an optional `tables` parameter to the Database constructor would make it better.  Or @andreibondarev can add that when doing https://github.com/andreibondarev/langchainrb/issues/129.  Let me know.

**Comment by mattlindsey** [comment]
Closing this to avoid too many open PRs, but leaving issue #138 open.  This can be done later if desired.

**Comment by andreibondarev** [comment]
@mattlindsey This is great. I tested it out locally:
```ruby
"CREATE TABLE schema_migrations(\n" +
"version string PRIMARY KEY\n" +
");\n" +
"CREATE TABLE ar_internal_metadata(\n" +
"key string PRIMARY KEY,\n" +
"value string,\n" +
"created_at datetime,\n" +
"updated_at datetime\n" +
");\n" +
"CREATE TABLE friendly_id_slugs(\n" +
"id integer PRIMARY KEY,\n" +
"slug string,\n" +
"sluggable_id integer,\n" +
"sluggable_type string,\n" +
"scope string,\n" +
"created_at datetime\n" +
");\n" +

...
```

I noticed that the foreign key statements are sitting outside of `CREATE TABLE ...()` declarations:
```ruby
...
");\n" +
"ALTER TABLE resources ADD FOREIGN KEY (user_id) REFERENCES users(id);\n" +
"CREATE TABLE active_storage_attachments(\n" +
...
```

Do you think we could put them inside? The way it's done in this paper again: https://arxiv.org/pdf/2204.00498.pdf (Page 11, Figure 8):
image

This 1 change, a spec and I think it's good to be merged!

andreibondarev:
This 1 change, a spec and I think it's good to be merged!

mattlindsey:
> This 1 change, a spec and I think it's good to be merged!

Great, but defining the foreign key after the table is perfectly valid and how I usually see it done, so I'd rather leave it for now.

I'd rather work with the open PR, so I'll add the spec there, and could you comment on that one. Thanks.

andreibondarev:
@mattlindsey I agree that it's valid as is, but I just assume since a published research paper claim to have effectiveness structuring db statement a certain -- it must be a good way to do it.

mattlindsey:
@andreibondarev I can change that, but could you put that comment on the newer PR.  This PR doesn't really solve much as the llm seems to understand even the ugly current dump.  
If you don't like the solution to the context size problem in the newer PR I have no problem with you updating my 138_schema branch or taking any of it for your own PR.

Issue #144, opened 2023-06-07
Add integration and/or system tests


mattlindsey:
I think we need the ability to add and run some 'Integration' tests that exercise interactions in high level components and use actual apis and keys.  They would be run only on request and could be run before each release.  

Start with a simple question to ChainOfThought with openai like in the README, with expectation that the result should be similar but not exactly equal to the result given in the README, since I assume the ai can respond slightly differently each time the test is called.

mattlindsey:
I'm going to try implementing a simple https://cucumber.io/ test.   It might work well here, but if it doesn't add value we don't have to use it.

andreibondarev:
@mattlindsey Do you envision that this would actually run in CI?

I'm also struggling a bit figuring out what value these feature tests would bring to this library?

mattlindsey:
If you run them in CI I think you'd catch errors sooner, like I think there's a gem dependency error now. (Might be wrong.)
Also, the agents are fairly high level, so testing interaction with other things using 'integration' testing is certainly necessary somewhere, I think.

andreibondarev:
I hope @technicalpickles doesn't mind that I pull him in. There was a mention of executing Jupyter notebooks or README code snippets in Discord. Would you have to have any thoughts here?

mattlindsey:
Also see where I implemented a couple of tests here to give a better idea: https://github.com/andreibondarev/langchainrb/pull/145

mattlindsey:
And for a wider range of testing it would be good if someone implemented Langchain::LLM::HuggingFace#complete.

technicalpickles:
> Start with a simple question to ChainOfThought with openai like in the README, with expectation that the result should be similar but not exactly equal to the result given in the README, since I assume the ai can respond slightly differently each time the test is called.

I was doing a course on deeplearning.ai, it was talking about how if you set a `temperature=0`, you should get the same results. The course was taught using jupyter notebooks, and the results they got doing the exercise matched what the AI was returning when I did them in the notebooks. I think it can be considered relatively stable?

> There was a mention of executing Jupyter notebooks or README code snippets in Discord. Would you have to have any thoughts here?

Yep! Here is what I suggested:

I've been thinking about getting the code in the README and in examples to be run as part of CI. did something like that for openfeature-sdk (https://github.com/open-feature/ruby-sdk/pull/40) ... I think the challenge for the README is making sure the fragment is complete enough to run, as well as having the right environment variables to make the call.

---

[...]

PR #145, opened 2023-06-07
Rspec 'integration' tests for optional use by contributors and ci

mattlindsey:
This isn't done, but I'm looking for feedback on using either cucumber or rspec for integration tests, which would aid in BDD development and be run before releases.   If  rspec is chosen, we can configure it to skip the 'integration' folder unless requested.

This PR includes both options for comparison.

For #144

One difficulty here is I am getting several different answers for the tests, so what's the acceptance criteria?  Just absence of errors?
5,845 soccer fields
204 soccer fields
4,125 soccer fields


alchaplinsky:
@mattlindsey Isn't RSpec enough for testing library code? Why do we need `cucumber` it doesn't seem to bring additional value to what we can already do with rspec testing an Agent.

mattlindsey:
@alchaplinsky Yeah.  I'll redo this proposal to only include the rspec file, exclude the 'integrations' folder when running rspec, and add a short note in the README suggesting contributors utilize the specs directly.

mattlindsey:
@alchaplinsky.  Is this better?  I think they're useful.

alchaplinsky:
@mattlindsey I think integration tests for concept like Agent is a good idea. But we need to mock requests to external APIs. I don't think that anyone would want to pay for tokens used during test runs.

mattlindsey:
Personally I wanted these to aid in development, but I can just keep them for myself. Existing tests already do the mocking.

mattlindsey:
@alchaplinsky Sorry, I meant to tag you on the last comment.
I added a line in the README to clarify the intent of the directory and files.

alchaplinsky:
Ok, got it. So this is a type of test that actually runs against real APIs and makes sure that code works with real integrations. Thus we skip running them every time when we run a test suite and we can enable them by setting the `INTEGRATION` env variable. Is that right? I'd probably name that variable something more descriptive like:
`INTEGRATION_TESTING = true`
`INTEGRATION_TESTS_ENABLED = true` 

Otherwise, it is not entirely clear what `INTEGRATION` stands for.

mattlindsey:
> `INTEGRATION_TESTS_ENABLED = true`

@alchaplinsky Yeah. That's better. Done

alchaplinsky (approved):

mattlindsey:
Closing this to avoid too many open PRs but leaving issue #144 open.

spec/integration/chain_of_thought_integration_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    openai = Langchain::LLM::OpenAI.new(api_key: ENV["OPENAI_API_KEY"])
+
+    agent = Langchain::Agent::ChainOfThoughtAgent.new(
+      llm: openai,
+      tools: [search_tool, calculator_tool]
+    )
+    result = agent.run(question: question)
+
+    expect(result).to include("distance between NYC and DC")
```
andreibondarev, line 25: One way that people do this is indexing into a vectorsearch DB all of the variations of possible acceptable answers and then run a vector search agains them to see if they "match".

PR #149, opened 2023-06-08
use llm with calculator instead of search

bborn:
Fallback to using the LLM to compose an expression the calculator can execute (rather than using search)

### DIFF
--- a/lib/langchain/tool/calculator.rb
+++ b/lib/langchain/tool/calculator.rb
@@ -2,6 +2,7 @@
 
 module Langchain::Tool
   class Calculator < Base
+    attr_reader :llm
     #
     # A calculator tool that falls back to the Google calculator widget
     #
@@ -18,27 +19,50 @@ module Langchain::Tool
       The input to this tool should be a valid mathematical expression that could be executed by a simple calculator.
     DESC
 
-    def initialize
+    def initialize(llm: nil)
       depends_on "eqn"
       require "eqn"
+
+      @llm = llm
     end
 
     # Evaluates a pure math expression or if equation contains non-math characters (e.g.: "12F in Celsius") then
-    # it uses the google search calculator to evaluate the expression
+    # it uses the llm to generate a math expression from the input and then evaluates it.
     # @param input [String] math expression
     # @return [String] Answer
-    def execute(input:)
+    def execute(input:, count: 0)
+      raise "Too many attempts" if count > 5
       Langchain.logger.info("[#{self.class.name}]".light_blue + ": Executing \"#{input}\"")
 
       Eqn::Calculator.calc(input)
-    rescue Eqn::ParseError, Eqn::NoVariableValueError
-      # Sometimes the input is not a pure math expression, e.g: "12F in Celsius"
-      # We can use the google answer box to evaluate this expression
-      # TODO: Figure out to find a better way to evaluate these language expressions.
-      hash_results = Langchain::Tool::SerpApi
-        .new(api_key: ENV["SERPAPI_API_KEY"])
-        .execute_search(input: input)
-      hash_results.dig(:answer_box, :to)
+    rescue Eqn::ParseError, Eqn::NoVariableValueError => e
+      if llm
+        output = llm.complete(
+          prompt: PROMPT_TEMPLATE % {question: input}
+        )
+        expression = output.strip.match(/```text(.*)```/m)[1].strip
+        execute(input: expression, count: count + 1)
+      else
+        raise e
+      end
     end
   end
 end
+
+PROMPT_TEMPLATE = "" "Translate a math problem into a expression that can be executed using Ruby's Eqn library. Use the output of running this code to answer the question.
+
+Question: ${{Question with math problem.}}
+```text
+${{single line mathematical expression that solves the problem}}
+```
+
+Begin.
+
+Question: What is 37593 * 67?
+
+```text
+37593 * 67
+```
+
+Question: %{question}
+" ""
--- a/spec/langchain/tool/calculator_spec.rb
+++ b/spec/langchain/tool/calculator_spec.rb
@@ -3,19 +3,21 @@
 require "eqn"
 
 RSpec.describe Langchain::Tool::Calculator do
+  subject {
+    llm = Langchain::LLM::OpenAI.new(api_key: "123")
+    described_class.new(llm: llm)
+  }
+
   describe "#execute" do
     it "calculates the result" do
       expect(subject.execute(input: "2+2")).to eq(4)
     end
 
-    it "calls Serp API when eqn throws an error" do
-      allow(Eqn::Calculator).to receive(:calc).and_raise(Eqn::ParseError)
+    it "calls the llm when eqn throws an error" do
+      allow_any_instance_of(Langchain::LLM::OpenAI).to receive(:complete)
+          .and_return("```text\n2+2```")
 
-      allow_any_instance_of(Langchain::Tool::SerpApi).to receive(:execute_search)
-        .with(input: "2+2")
-        .and_return({answer_box: {to: 4}})
-
-      subject.execute(input: "2+2")
+      subject.execute(input: "what is 2+2?")
     end
   end
 


mattlindsey:
Does ChainOfThought still work after this?  If you pass it the same llm object?

mattlindsey:
It seems to get confused and call the llm on a valid equation:

```
3.2.1 :012 > result = agent.run(question:  "How many full soccer fields would be needed to cover the distance between NYC and DC in a straight line?")
[LangChain.rb] [Langchain::Agent::ChainOfThoughtAgent]: Sending the prompt to the Langchain::LLM::OpenAI LLM
[LangChain.rb] [Langchain::Agent::ChainOfThoughtAgent]: Invoking "Langchain::Tool::Calculator" Tool with "(360 ft * 160 ft) / (393 miles)"
[LangChain.rb] [Langchain::Tool::Calculator]: Executing "(360 ft * 160 ft) / (393 miles)"
[LangChain.rb] [Langchain::Tool::Calculator]: Executing "(360 * 160) / 393"
[LangChain.rb] [Langchain::Agent::ChainOfThoughtAgent]: Sending the prompt to the Langchain::LLM::OpenAI LLM
[LangChain.rb] [Langchain::Agent::ChainOfThoughtAgent]: Invoking "Langchain::Tool::Calculator" Tool with "146.5648854961832 * 0.000000386102159"
[LangChain.rb] [Langchain::Tool::Calculator]: Executing "146.5648854961832 * 0.000000386102159"
[LangChain.rb] [Langchain::Agent::ChainOfThoughtAgent]: Sending the prompt to the Langchain::LLM::OpenAI LLM
[LangChain.rb] [Langchain::Agent::ChainOfThoughtAgent]: Invoking "Langchain::Tool::Calculator" Tool with "5.6589018723664115e-05 / (360 ft * 160 ft)"
[LangChain.rb] [Langchain::Tool::Calculator]: Executing "5.6589018723664115e-05 / (360 ft * 160 ft)"
/Users/mattlindsey/github/langchainrb/lib/langchain/tool/calculator.rb:43:in `rescue in execute': undefined method `[]' for nil:NilClass (NoMethodError)

        expression = output.strip.match(/```text(.*)```/m)[1].strip
                                                          ^^^
	from /Users/mattlindsey/github/langchainrb/lib/langchain/tool/calculator.rb:33:in `execute'

File: lib/langchain/tool/calculator.rb [hunk 1]
```diff
@@ -18,27 +19,50 @@ class Calculator < Base
...
-      # Sometimes the input is not a pure math expression, e.g: "12F in Celsius"
-      # We can use the google answer box to evaluate this expression
-      # TODO: Figure out to find a better way to evaluate these language expressions.
-      hash_results = Langchain::Tool::SerpApi
-        .new(api_key: ENV["SERPAPI_API_KEY"])
-        .execute_search(input: input)
-      hash_results.dig(:answer_box, :to)
+    rescue Eqn::ParseError, Eqn::NoVariableValueError => e
+      if llm
+        output = llm.complete(
```
alchaplinsky, line 40: I think passing and using LLM within calculator tool is an overkill and not exactly correct separation of concerns: The idea of a `tool` is that it is a simple and straightforward class that is responsible for doing only 1 thing.
`Agent` on the other hand is a more complex component that is able to manipulate tools and LLMs to get to the end goal. Therefore I think Agent should be responsible for making a decision whether we should make a call to LLM or Search if a calculator fails.

Otherwise, why would we have this separation into`Agent`,`Search`, `Calculator` tools if any of them can use LLM and execute other tools to perform a certain task?

lib/langchain/tool/calculator.rb
```diff
@@ -18,27 +19,50 @@ class Calculator < Base
       The input to this tool should be a valid mathematical expression that could be executed by a simple calculator.
     DESC
 
-    def initialize
+    def initialize(llm: nil)
```
alchaplinsky, line 22: I'd instead allow to pass and option what to do in case of error (like an error callback). Which would come from an Agent, because agent should decide what to do if a certain tool fails.

lib/langchain/tool/calculator.rb
bborn: Disagree here - I think the tool should be called "LlmCalculator" and it  _should_ require passing in an LLM. The point is that it's a tool that can take natural language input and do calculations with it. If it's just a light wrapper around `Eqn`, what's the point? 

Or perhaps it needs to be thought of as more of a `Chain` than a `Tool`, which might make sense (that's exactly how Langchain.py treats it: https://python.langchain.com/en/latest/modules/chains/examples/llm_math.html).

That would be fine with me, if you think it makes more sense. But the bottom line for me is that Agent tools/chains/whatever will, in many cases, need access to the LLM that's driving the agent.

lib/langchain/tool/calculator.rb
bborn: See my comment above: I see what you're getting at, but my feeling is a tool that simply takes input and passes it to `Eqn` is not really worth the trouble.

alchaplinsky: > A tool that simply takes input and passes it to Eqn is not really worth the trouble

It is. The tool is meant to perform a single task. And all tools need to have a unified API, so that agent knows how to call then, pass an input and get a result. That is the purpose of the tool. Whether calculator calls `Eqn` and returns a result or does a number of other calculations that is implementation details which can change. However, in order for Agent to do a calculation you need to give it a Calculator tool with a standardized interface.

lib/langchain/tool/calculator.rb
```diff
@@ -18,27 +19,50 @@ class Calculator < Base
...
       Eqn::Calculator.calc(input)
-    rescue Eqn::ParseError, Eqn::NoVariableValueError
-      # Sometimes the input is not a pure math expression, e.g: "12F in Celsius"
-      # We can use the google answer box to evaluate this expression
-      # TODO: Figure out to find a better way to evaluate these language expressions.
-      hash_results = Langchain::Tool::SerpApi
-        .new(api_key: ENV["SERPAPI_API_KEY"])
-        .execute_search(input: input)
-      hash_results.dig(:answer_box, :to)
+    rescue Eqn::ParseError, Eqn::NoVariableValueError => e
```
andreibondarev, line 38: @bborn I'm really curious -- what if the response from the Tool, that gets passed back as the `Observation:` to the Agent/to the LLM is "Unable to calculate "#{input}" due to #{error}. Please reformat the input.". Could this trigger the ChainOfThought to **actually** call the Calculator tool again with re-formatted input?

bborn: @andreibondarev I like it. Worth trying to see how it works. But you might be using the Calculator tool/chain/etc in a different context (e.g. a Sequential agent rather than a Chain of Thought). So ultimately I think it's the Calculator's job to take any reasonable input that is asking a math-like question, and try to answer it. I suppose if the input is e.g. "What is a dolphin?" then the Calculator should say "That's not a math question", but otherwise, I want it to turn "what's the area of a circle with radius of 4? and then divide that number by three" into "(π(4^2))/3".

lib/langchain/tool/calculator.rb
mattlindsey: @alchaplinsky Great point about the different components.  @andreibondarev Worth adding the vision for these here https://github.com/andreibondarev/langchainrb/wiki#concepts?

lib/langchain/tool/calculator.rb
andreibondarev: Here's another idea -- in the Tool's description providing it a few examples, something along the lines of:
```ruby
      Useful for getting the result of a math expression.
      The input to this tool should be a valid mathematical expression that could be executed by a simple calculator.
      Usage:
        Action Input: 1+1
        Calculator: 2
        ...
```

bborn:
@andreibondarev still having issues with this tool as currently defined. I get failures when running because it requires `ENV["SERPAPI_API_KEY"]` to be set, and I'm not using SerpAPI. I don't think any of the tools should require ENV vars in order to work. 

My vote would be to use the LLM to deal with calculator errors because 1) that's what Langchain does and 2) it's reasonable that tools would have access to an LLM (or in the case of Agents, to the LLM that the Agent uses).

I don't think it makes sense to push that back up to the Agent (there may be lots of different types of Agents and not all will be suited to figuring out the right way to call `Eqn::Calculator`). I think it's the Calculator tool's job to convert natural language into math expressions, and return the result. If that makes this tool more of an Agent, that's fine with me too.

If you don't want to do that, and the tool is really just meant to be a simple interface to `Eqn::Calculator` then it also should NOT require falling back to another tool (`Langchain::Tool::GoogleSearch`) and depend on an ENV var.

mattlindsey:
@bborn `Equation` tool with no external API requirements and more autonomous `Calculator` agent?

andreibondarev:
> If you don't want to do that, and the tool is really just meant to be a simple interface to Eqn::Calculator then it also should NOT require falling back to another tool (Langchain::Tool::GoogleSearch) and depend on an ENV var.

@bborn I'm going to take a stab at this ^^

andreibondarev:
@bborn @mattlindsey Closing this PR and merging https://github.com/andreibondarev/langchainrb/pull/197 for now. I'm those improvements can steer the LLM to retrying the calculator tool whenever the input is not a valid math equation.

PR #151, opened 2023-06-08
Fix variable names in readme example

breckenedge:


andreibondarev (approved):
Nice catch! And thank you for your contribution @breckenedge !

PR #152, opened 2023-06-08
Yard doc improvements

technicalpickles:
I was chatting with @andreibondarev about including additional files in the YARD output. You could imagine moving the wiki into a docs folder, for example.

 I started doing that, by including all markdown files, the license, and example files... and then it kinda got out from under me 🤪 

I took a bunch of inspiration from how Rails organizes it's docs. For each "component", they put the docs on how to use them onto the Base class. So I did that. Here's a grab bag of changes:

- make sure the documentation is on the right constant (several were on the `Langchain::<Component>` for example)
- Add links to the LLM providers when possible
- Add the correct signature to LLM::Base methods. Document that they can pass in params to the client, and can raise notimplemented error if the LLM doesn't provide it
- changed triple backtick fencing to 4 spaces. it automatically renders code as ruby syntax

I kinda ran out of steam though, but figured this was a good spot to stop for feedback. I think there is some open questions about what type of docs go where, but probably better to discuss on https://github.com/andreibondarev/langchainrb/issues/54

sidenote: It's also worth noting... YARD syntax is different than Markdown, and slightly different than rdoc.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,4 +61,4 @@ jobs:
           bundler: default
           bundler-cache: true
       - name: Build docs
-        run: bundle exec rake yard
+        run: bundle exec yardoc
--- a/.yardopts
+++ b/.yardopts
@@ -1,0 +1,1 @@
+--fail-on-warning lib/**/*.rb - **/*.md LICENSE.txt
--- a/.yardopts
+++ b/.yardopts
@@ -1,1 +1,1 @@
---fail-on-warning lib/**/*.rb - **/*.md LICENSE.txt
+--fail-on-warning lib/**/*.rb - **/*.md examples/**/*.rb LICENSE.txt
--- a/.yardopts
+++ b/.yardopts
@@ -1,1 +1,0 @@
---fail-on-warning lib/**/*.rb - **/*.md examples/**/*.rb LICENSE.txt
--- a/Rakefile
+++ b/Rakefile
@@ -14,5 +14,4 @@ Rake::Task["spec"].enhance do
 end
 
 YARD::Rake::YardocTask.new do |t|
-  t.options = ["--fail-on-warning"]
 end
--- a/lib/langchain.rb
+++ b/lib/langchain.rb
@@ -38,7 +38,6 @@ module Langchain
     autoload :Database, "langchain/tool/database"
   end
 
-  # Processors load and parse/process various data types such as CSVs, PDFs, Word documents, HTML pages, and others.
   module Processors
     autoload :Base, "langchain/processors/base"
     autoload :CSV, "langchain/processors/csv"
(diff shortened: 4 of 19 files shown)


andreibondarev:
@technicalpickles Wow. Legend.

I hope you don't mind me committing directly to your branch?

technicalpickles:
@andreibondarev nope, not at all!

andreibondarev:
> Add the correct signature to LLM::Base methods. Document that they can pass in params to the client, and can raise notimplemented error if the LLM doesn't provide it

@technicalpickles How do you think we should best handle slightly different method signatures? Of course we're trying to create a common DSL for all of the `Vectorsearch::*` subclasses but there will be minor differences. For example the Chroma DB [initialize](https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/vectorsearch/chroma.rb#L19) does not accept an `api_key:` hence it was set to `api_key: nil`.

This is why the `Vectorsearch::Base` class used "argument forwarding".

technicalpickles:
Oh, hm, that's a good question! I'm not sure how to best handle unimplemented methods on the super class.

I think it makes sense to document `params` as keyword args for the base class, but then document any specific params used by the subclass.

---

Oh, and also I finished copying in the wiki. Credit for the content itself goes to whoever wrote those 😁

technicalpickles:
Until there's a consistent API, I think we can just use `...` in the super classes. I'll see about fixing those up and removing references to the params.

technicalpickles:
I'm not sure what is causing the docs failure yet, but I can reproduce it:

```
bundle install --path vendor/cache
bundle exec yardoc
```

andreibondarev:
> I'm not sure what is causing the docs failure yet, but I can reproduce it:
> 
> ```
> bundle install --path vendor/cache
> bundle exec yardoc
> ```

It works for me if I install the gems to system with just `bundle install`. Double-check your `.bundle/config` in the project directory.

```ruby
 -> langchain [yard-improvements*]: yard
Files:          42angchain/Agent/ChainOfThoughtAgent.htmltmllates.htmlagent.rb
Modules:         9 (    8 undocumented)
Classes:        42 (   27 undocumented)
Constants:      36 (   27 undocumented)
Attributes:     20 (    0 undocumented)
Methods:       137 (    9 undocumented)
 70.90% documented
 ```

andreibondarev:
@technicalpickles 🚀

andreibondarev (approved):

technicalpickles:
> It works for me if I install the gems to system with just bundle install. Double-check your .bundle/config in the project directory.

Right, it works locally without any changes. It was failing on CI though: https://github.com/andreibondarev/langchainrb/actions/runs/5227093582/jobs/9438385274#step:4:6

andreibondarev:
> > It works for me if I install the gems to system with just bundle install. Double-check your .bundle/config in the project directory.
> 
> 
> 
> Right, it works locally without any changes. It was failing on CI though: https://github.com/andreibondarev/langchainrb/actions/runs/5227093582/jobs/9438385274#step:4:6

Yeah I was thinking of just creating a separate branch with the docs/ folder checked in and then deploying it via GitHub Pages.

PR #154, opened 2023-06-09
🔍 Weather tool implementation

mattlindsey:
For current and forecast weather by city.  Historical weather seems to always cost, so not supporting that now.


lib/langchain/tool/weather.rb
```diff
@@ -0,0 +1,50 @@
...
+    # Gem requirements:
+    #   gem "open-weather-ruby-client", "~> 0.3.0"
+    #
+
+    NAME = "weather"
+
+    description <<~DESC
+      Useful for getting current, forecast, or historical weather data
+  
+      The input to this tool should be a city name followed by the type of weather you want to get.
```
andreibondarev, line 17: Just curious -- did we copy the description from somewhere else?

lib/langchain/tool/weather.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module Langchain::Tool
+  class Weather < Base
+    #
+    # A weather tool that gets current, forecast, or historical weather data
+    #
+    # Gem requirements:
+    #   gem "open-weather-ruby-client", "~> 0.3.0"
+    #
```
andreibondarev, line 10: Please write out Usage instructions. There's some examples of this in other files.

lib/langchain/tool/weather.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module Langchain::Tool
+  class Weather < Base
+    #
+    # A weather tool that gets current, forecast, or historical weather data
```
andreibondarev, line 6: Would you please a link to where the API key can be obtained from? Just for convenience.

mattlindsey:
Works for 'current' weather
image

mattlindsey:
@andreibondarev There are two TODOs in the code:
1:  Return weather forecast by looking up lat and lon of city, then calling client.one_call(lat: 33.441792, lon: -94.037689)
     May need another gem for that.
2. Return historical weather.  Also requires lat and lon, and will return error if not using a paid API Key.

Is it good enough to return just current weather for now, since these tools are experimental.  I can get to the TODOs in a couple days.

lib/langchain/tool/weather.rb
mattlindsey: done

lib/langchain/tool/weather.rb
mattlindsey: done

lib/langchain/tool/weather.rb
```diff
@@ -0,0 +1,73 @@
...
+    def execute(input:)
+      Langchain.logger.info("[#{self.class.name}]".light_blue + ": Executing for \"#{input}\"")
+
+      city = input.split(", ").first
+      if input.count(",") > 1
+        type = input.split(", ")[0..-2][1]
+        units = input.split(", ").last
+      else
+        type = input.split(", ").last
+      end
```
andreibondarev, line 54: I think you can simplify this whole block with:
```ruby
input_array = input.split(",")
city, type, units = *input_array
```

mattlindsey:
FYI:  geocoder gem can do it, but the tool would require Google Maps Geocoding API key:

require 'geocoder'

result = Geocoder.search("New York City").first
latitude = result.latitude
longitude = result.longitude
puts "Latitude: #{latitude}, Longitude: #{longitude}"

lib/langchain/tool/weather.rb
mattlindsey: Tried but that breaks it even though looks like it should work.  Will look into it.

mattlindsey: Fixed.  Will wait for further input on wisdom of adding additional gem and google maps api key requirement.

andreibondarev: @mattlindsey What do we need to pass the lat/long for? Does the API not accept the actual place string?

mattlindsey: No.  I believe the `forecast` data cannot be provided without lat/long, and the gem cannot determine those itself.  Weird situation.

mattlindsey: @andreibondarev To be clear, the API (not gem) provides the following with the free plan (1000 calls per day), but forecasts are by lat/long, so the Tool could do the following for free if the consumer of the tool passes in lat/long to `execute`:

Current weather
Minute forecast for 1 hour
Hourly forecast for 48 hours
Daily forecast for 8 days
Government weather alerts
Historical weather data for any timestamp for 40+ years historical archive

Taken from https://openweathermap.org/api
History daily aggregated weather data for 40+ years historical archive

PR #156, opened 2023-06-09
Chat message history support for LLM:OpenAI

alchaplinsky:
## Changes
- Allow passing message history (as a simple array) to `chat` method of LLM::OpenAI. 
- Keep prompt argument for now, so that messages could be passed in several ways:

with no message history:
```ruby
client.chat(prompt: "How are you?")
```

with prompt and message history:
```ruby
client.chat(prompt: 'How are you?', messages: [{role: "system", content: "You are a chatbot"}])
```

with only message history:
```ruby
client.chat(messages: [{role: "system", content: "You are a chatbot"}, { role: "user", content: "How are you?"}])
```

### DIFF
--- a/lib/langchain/llm/openai.rb
+++ b/lib/langchain/llm/openai.rb
@@ -103,7 +103,7 @@ module Langchain::LLM
     end
 
     def validate_max_tokens(messages, model)
-      Langchain::Utils::TokenLengthValidator.validate_max_tokens!(messages.map { |m| m[:content] }, model)
+      Langchain::Utils::TokenLengthValidator.validate_max_tokens!(messages, model)
     end
   end
 end
--- a/lib/langchain/llm/openai.rb
+++ b/lib/langchain/llm/openai.rb
@@ -61,13 +61,17 @@ module Langchain::LLM
     # Generate a chat completion for a given prompt
     #
     # @param prompt [String] The prompt to generate a chat completion for
+    # @param messages [Array] The messages that have been sent in the conversation
     # @return [String] The chat completion
     #
-    def chat(prompt:, **params)
-      parameters = compose_parameters DEFAULTS[:chat_completion_model_name], params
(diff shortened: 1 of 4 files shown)


lib/langchain/llm/openai.rb
```diff
@@ -61,13 +61,17 @@ def complete(prompt:, **params)
     # Generate a chat completion for a given prompt
     #
     # @param prompt [String] The prompt to generate a chat completion for
+    # @param messages [Array] The messages that have been sent in the conversation
     # @return [String] The chat completion
     #
-    def chat(prompt:, **params)
-      parameters = compose_parameters DEFAULTS[:chat_completion_model_name], params
+    def chat(prompt: "", messages: [], **params)
+      return if prompt.empty? && messages.empty?
```
andreibondarev, line 68: Should we raise an error? Returning `nil` would be rather confusing to a developer.

lib/langchain/utils/token_length_validator.rb
```diff
@@ -51,13 +51,14 @@ class TokenLengthValidator
...
-      # @param text [String] The text to validate
+      # @param content [String | Array<String>] The text or array of texts to validate
       # @param model_name [String] The model name to validate against
       # @return [Integer] Whether the text is valid or not
       # @raise [TokenLimitExceeded] If the text is too long
       #
-      def self.validate_max_tokens!(text, model_name)
-        text_token_length = token_length(text, model_name)
+      def self.validate_max_tokens!(content, model_name)
+        text_token_length = Array(content).sum { |text| token_length(text, model_name) }
```
andreibondarev, line 60: Are there any other use-cases where `validate_max_tokens!()` would be called with content being an array?

lib/langchain/llm/openai.rb
alchaplinsky: yeah, an error is better

andreibondarev (approved):

PR #160, opened 2023-06-10
Add basic Chat implementation

alchaplinsky:
## Motivation
An intuitive and unified API for creating chats with different LLMs.

## Example
```ruby
llm = Langchain::LLM::OpenAI.new api_key: ENV['OPEN_AI_KEY']
chat = Langchain::Chat.new llm: llm

chat.set_context "You are a travel assistant that only responds to travel questions. And responds ironically to other topics"

chat.message "What is most complex math problem?"
 => "Oh, I'm sorry, I thought you were looking for travel advice, not trying to give me a headache with math problems. But if you insist, the most complex math problem is probably the one that involves trying to calculate how much money you'll have left after booking a trip without checking your bank account first."

chat.message 'What is the second most visited place in Thailand?'
=> "Ah, a travel question! The second most visited place in Thailand after Bangkok is probably Phuket. It's a beautiful island with stunning beaches, great nightlife, and plenty of activities to keep you entertained. Just be prepared for the crowds of tourists and the occasional overpriced souvenir."
```

## Todo
- [ ] Keep general `Langchain::Chat` module with standardised `messages` format. Re-map messages to provider specific structure:
  - **OpenAi**: `{role: "assistant", content: "Some text..." }`
  - **Google PaLM2**: `{author: "ai", content: "Some text..."}`
- [ ] Allow to pass options and change them. Like changing `temperature`, `top_k` etc. 
- [ ] Calculate the number of tokens left after each message and truncate history as needed to keep conversation going instead of running into token limit exceeded.

### DIFF
--- a/lib/langchain/chat.rb
+++ b/lib/langchain/chat.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+module Langchain
+  class Chat
+    attr_reader :context
+
+    def initialize(llm:, **options)
+      @llm = llm
+      @context = nil
+      @examples = []
+      @messages = []
+    end
+
+    # Set the context of the conversation. Usually used to set the model's persona.
+    # @param message [String] The context of the conversation
+    def set_context(message)
+      @context = message
+    end
+
+    # Add examples to the conversation. Used to give the model a sense of the conversation.
+    # @param examples [Array<Hash>] The examples to add to the conversation
+    def add_examples(examples)
+      @examples.concat examples
+    end
+
+    # Message the model with a prompt and return the response.
+    # @param message [String] The prompt to message the model with
+    # @return [String] The response from the model
+    def message(message)
+      build_history if @messages.empty?
+      append_user_message(message)
+      response = llm_response
+      append_ai_message(response)
+      response
+    end
+
+    private
+
+    def llm_response
+      @llm.chat(messages: @messages)
+    end
+
+    def build_history
+      set_system_message(@context) if !!@context
+      @messages.concat @examples
+    end
+
+    def append_ai_message(message)
+      @messages << {role: "assistant", content: message}
+    end
+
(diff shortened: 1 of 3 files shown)


lib/langchain/chat.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    private
+
+    def llm_response
+      @llm.chat(messages: @messages)
+    end
+
+    def build_history
+      set_system_message(@context) if !!@context
+      @messages.concat @examples
```
andreibondarev, line 45: For OpenAI, do the examples get stuffed into the same `messages:` array but as `{ role: "system", content: "..."}`?

lib/langchain/chat.rb
```diff
@@ -0,0 +1,60 @@
...
+      append_user_message(message)
+      response = llm_response
+      append_ai_message(response)
+      response
+    end
+
+    private
+
+    def llm_response
+      @llm.chat(messages: @messages)
```
andreibondarev, line 40: We may need to restructure it for Google Palm, since Google Palm accepts `examples:`, `context:`, and `messages:` as separate parameters.

andreibondarev (commented):
Really solid abstraction thus far! 🚀

lib/langchain/chat.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+module Langchain
+  class Chat
```
andreibondarev, line 4: What are your thought of calling this class something like `ChatSession`? I don't have any strong feeling one way or another here.

lib/langchain/chat.rb
alchaplinsky: Yeah, so my idea is that this class will accept a unified standard input. And then this input will be transformed to the format accepted by LLM (added that to TODO).

PR #163, opened 2023-06-10
Case insensitive tool lookup

bborn:
LLM sometimes doesn't get the case right when selecting a tool. Switch to case insensitive matching.


lib/langchain/agent/chain_of_thought_agent/chain_of_thought_agent.rb
```diff
@@ -50,8 +51,14 @@ def run(question:)
         question: question,
         tools: tools
       )
+      Langchain.logger.info("[#{self.class.name}]".light_blue + ": Agent prompt is\n#{prompt}")
 
+      iterations = 0
       loop do
+        if (iterations + 1) > max_iterations
+          break "Agent stopped after #{iterations} iterations"
```
alchaplinsky, line 59: 👍

lib/langchain/agent/chain_of_thought_agent/chain_of_thought_agent.rb
```diff
@@ -67,18 +74,24 @@ def run(question:)
...
+            Langchain.logger.info("[#{self.class.name}]".red + ": Invoking \"#{tool.class}\" Tool with \"#{action_input}\"")
 
-          # Call `execute` with action_input as the input
-          result = tool.execute(input: action_input)
+            # Call `execute` with action_input as the input
+            result = tool.execute(input: action_input)
+          else
+            Langchain.logger.info("[#{self.class.name}]".red + ": Tool \"#{action}\" not found.")
+
+            result = "Tool \"#{action}\" not found. I need to try a different tool."
```
alchaplinsky, line 85: Wondering if this should be an exception. Since developer is responsible for passing a list of supported tools, maybe we should error out on them instead of returning a string result.

bborn: I don't follow ... the issue here is that the LLM often picks the wrong tool name. The result here is meant to get passed back into the chain of thought loop so the LLM can try again and pick a different (hopefully correct) tool name. Raising an error here would break the loop, so I don't think that's what we want.

alchaplinsky: Hm... I thought LLM should only pick the tool from a given list that we [pass within the prompt](https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/agent/chain_of_thought_agent/chain_of_thought_agent_prompt.json#L3). And the issue addressed here is that a developer passes some tools that are non-existing in the codebase.

If that's the case that LLM returns the wrong tool. Maybe we should also try improving the prompt so that doesn't happen.

bborn: Agreed about trying to improve the prompt. But even though the allowed tool list is passed  in the prompt, the LLM will sometimes return a completion with an incorrect tool choice.

technicalpickles: This was from before OpenAI announced function support, right? Maybe that would help with getting more consistent calls?

bborn: Good point, but we're trying to support more than just OAI LLMs, right?

andreibondarev:
Closing this PR since Agents will be replace with Assistants that utilize Function Calling.

PR #164, opened 2023-06-10
Change SerpApi class name to Search

mattlindsey:
Change LangchainTool::SerpApi to Langchain::Tool::Search.

Not necessary, but SerpApi seems confusing and unlike other tool class names.


mattlindsey:
Change filename too?  langchain/tool/serp_api.rb

mattlindsey:
Changed filename too. Works:

image

alchaplinsky (approved):
I like it. The name is more general. Underlying API change over time but tool's purpose and the name stays the same.

andreibondarev:
Solid. The only idea I might throw out there is whether this tool should actually be name "google_search"? Since it utilizes a 3rd party SaaS solution that executes Google searches. I don't have a super strong feeling here either way.

mattlindsey:
Cool.  Changed name to google_search and ran a few tests.  Looks good.

mattlindsey:
@andreibondarev No problem, but spec is failing due to GoogleSearch already defined.  Can you fix, or advise on fix?

mattlindsey:
> Was it a bad merge?

@andreibondarev I think it disabled tests running just `rspec`

andreibondarev:
The build is green on the main branch though.

mattlindsey:
@andreibondarev 
Yeah...
The logging PR only runs tests locally when marked fit(for me at least)

andreibondarev:
@rickychilcott @mattlindsey Fixed, thank you!

bborn:
@andreibondarev @mattlindsey  I know this is already merged, but can we reconsider? I want to add a search tool that uses the CSE API (https://developers.google.com/custom-search/v1/introduction), so I think this naming will be confusing in that case. I think this tool should either be called `SerpApi` (because that's the API is uses), or it should stay `GoogleSearch` but allow for selecting multiple API providers (SerpAPI, CSE, [ScrapingBee](https://www.scrapingbee.com/documentation/google/), [Serper](https://www.google.com/search?client=safari&rls=en&q=google+serper&ie=UTF-8&oe=UTF-8), etc.)

mattlindsey:
I'm OK with either, but then I should have named `Weather` by it's API too: `OpenWeatherApi`, since there are several.

bborn:
@mattlindsey  yeah - good point. If we're worried about having a tool class for every different API, then I think we need to generalize and allow swapping in different API providers in those cases. In particular, for Google Search, feels important, because SerpAPI is $$ and other options are free/cheaper, so it seems likely people will want choices.

mattlindsey:
@bborn And it will affect agents, since passing in 1 tool vs many to the llm would operate differently.  For 1 tool with many APIs, the tool description would have to include the different API choices that are enabled.  Right?

alchaplinsky:
Agree with @bborn. There should be 1 tool per task that it should perform. So, generic names like `Search`, `Calculator`, `Calendar`, `Weather`, etc. make sense. Which API is used is internal implementation which should not be a part of the tool name. Though, this implementation should be configurable through options passed to the tool. 

So that `Search` tool can use Serp API or any other API or even perform a search on duckduckgo web and then parse the results page. But all of this should be encapsulated in a generic `Search` tool with a unified tool interface.

PR #168, opened 2023-06-11
Rename Langchain::Chat to Langchain::Conversation

alchaplinsky:
### Changes
Renaming of a newly introduced API: `Chat` => `Conversation`. I like the meaning of "conversation" more. However, it is a longer name and more characters to type. 

Let's vote for this renaming 👍  or 👎

### DIFF
--- a/lib/langchain/conversation.rb
+++ b/lib/langchain/conversation.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Langchain
-  class Chat
+  class Conversation
     attr_reader :context
 
     def initialize(llm:, **options)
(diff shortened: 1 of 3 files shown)


lib/langchain/conversation.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Langchain
-  class Chat
+  class Conversation
```
andreibondarev, line 4: Think we can also annotate the code, write out what it's used for, what it does and usage examples?

andreibondarev (approved):

lib/langchain/conversation.rb
alchaplinsky: Adding in the next PR https://github.com/andreibondarev/langchainrb/pull/169

PR #171, opened 2023-06-12
Add namespace option to pgvector

bborn:
* Allow passing a namespace to pgvector client init
* Use exec_prepared for PG queries


lib/langchain/vectorsearch/pgvector.rb
```diff
@@ -18,23 +18,27 @@ class Pgvector < Base
...
     # @param index_name [String] The name of the table to use for the index
     # @param llm [Object] The LLM client to use
     # @param api_key [String] The API key for the Vectorsearch DB (not used for PostgreSQL)
-    def initialize(url:, index_name:, llm:, api_key: nil)
+    # @param namespace_column [String] The name of the column to use for the namespace
+    # @param namespace [String] The namespace to use for the index when inserting/querying
+    def initialize(url:, index_name:, llm:, api_key: nil, namespace_column: nil, namespace: nil)
       require "pg"
       require "pgvector"
+      require "openai"
```
andreibondarev, line 32: Why do we need to require the OpenAI here? What if you were intending to use a different LLM for embeddings.

lib/langchain/vectorsearch/pgvector.rb
```diff
@@ -18,23 +18,27 @@ class Pgvector < Base
...
 
       @client = ::PG.connect(url)
       registry = ::PG::BasicTypeRegistry.new.define_default_types
       ::Pgvector::PG.register_vector(registry)
       @client.type_map_for_results = PG::BasicTypeMapForResults.new(@client, registry: registry)
 
-      @index_name = index_name
-      @quoted_table_name = @client.quote_ident(index_name)
+      @table_name = index_name
```
andreibondarev, line 39: We still need the quote_ident() here.

bborn: Oh, my understanding was `prepare` takes care of that. Wrong?

lib/langchain/vectorsearch/pgvector.rb
bborn: I was getting a flaky errors (intermittent) about OpenAI not being defined when running tests.

lib/langchain/vectorsearch/pgvector.rb
```diff
@@ -45,11 +49,16 @@ def initialize(url:, index_name:, llm:, api_key: nil)
     # @return [PG::Result] The response from the database
     def add_texts(texts:)
       data = texts.flat_map do |text|
-        [text, llm.embed(text: text)]
+        [text, llm.embed(text: text), namespace]
```
andreibondarev, line 52: Would this blow up if the `namespace == nil`?

andreibondarev:
@bborn What are your thoughts on converting this pgvector integration to use [Squel](https://github.com/pgvector/pgvector-ruby#sequel)? 😇 

PG is turning out to be super low-level and I'm not sure we want to go all out and introduce ActiveRecord on the other side of the spectrum yet.

bborn:
I love it. I don't like write SQL anywhere. Only kept going that way because that was what was there.

andreibondarev:
The build is green on the main branch.

bborn:
@andreibondarev can you take a look here? Tests keep failing on CI, while locally they pass about 50% of the time. No clue why the flakiness.

andreibondarev:
@bborn The PR looks good on paper, I'd like to just test it a bit more by running the whole end-to-end locally. Need a bit more time.

lib/langchain/vectorsearch/pgvector.rb
```diff
@@ -13,60 +13,70 @@ class Pgvector < Base
...
-      client.exec(
-        <<~SQL
-          CREATE TABLE IF NOT EXISTS #{quoted_table_name} (
-            id serial PRIMARY KEY,
-            content TEXT,
-            vectors VECTOR(#{default_dimension})
-          );
-        SQL
-      )
+      db.run "CREATE EXTENSION IF NOT EXISTS vector"
```
andreibondarev, line 71: This throws an error:
```ruby
Sequel::DatabaseError (PG::UndefinedFile: ERROR:  could not open extension control file "/Applications/Postgres.app/Contents/Versions/13/share/postgresql/extension/vector.control": No such file or directory)
```

andreibondarev: Does it work for you? 
``ruby
pgvector = Langchain::Vectorsearch::Pgvector.new url: url, llm: llm, index_name: "vectors"
pgvector.create_default_schema
```

Comment (bborn) [reply lib/langchain/vectorsearch/pgvector.rb:71]: Do you have pgvector installed? Looks like it's not there for pg@13?

Comment (bborn) [reply lib/langchain/vectorsearch/pgvector.rb:71]: Yes, works for me.

File: lib/langchain/vectorsearch/pgvector.rb [hunk 5]
```diff
@@ -8,65 +8,74 @@ class Pgvector < Base
...
 
     # @param url [String] The URL of the PostgreSQL database
     # @param index_name [String] The name of the table to use for the index
     # @param llm [Object] The LLM client to use
-    # @param api_key [String] The API key for the Vectorsearch DB (not used for PostgreSQL)
-    def initialize(url:, index_name:, llm:, api_key: nil)
-      require "pg"
+    # @param namespace_column [String] The name of the column to use for the namespace
+    # @param namespace [String] The namespace to use for the index when inserting/querying
+    def initialize(url:, index_name:, llm:, namespace_column: nil, namespace: nil)
```
andreibondarev, line 28: @bborn Btw -- what are `namespace_column` and `namespace` used for in general?

bborn: My use case: in AgentHQ, I let people create indices from various loader sources (files, rss, intercom, etc.). I want to store all the vectors in the same table (rather than a table per index), so I need a way (namespace) of differentiating them. 

Pinecone does similar: https://docs.pinecone.io/docs/namespaces

bborn: Ah good point - yeah it's not necessary to make it configurable. Happy to make it hardcoded if you think that's better.

andreibondarev: Yeah, I think so!

bborn:
@andreibondarev updated - made namespace column name hard coded. Still failing tests on CI (passing locally: https://d.pr/i/bNKNz7)

bborn:
@andreibondarev updates here?

bborn:
@andreibondarev bump.

andreibondarev:
Hey @bborn ! I merge [this PR](https://github.com/andreibondarev/langchainrb/pull/248) first because I figured it was going to be easier for you to resolve the merge conflicts here vs the other way around.

Also -- do you have any clue why the specs are failing? I'm okay with merge this PR after these changes/fixes though.

bborn:
@andreibondarev updated (using Sequel for upserts) and tests passing (at least locally)

andreibondarev:
@bborn Hmm, yeah, I've got the specs passing locally as well. Not too sure about why they're failing on CI.

bborn:
@andreibondarev yeah, weird. `method content= doesn't exist` could this be because the `products` table hasn't been created in CI, but locally it got created on a previous test run and never dropped?

andreibondarev:
@bborn I guess this could be easily tested by dropping the `products` table locally and then re-running the test suite.

andreibondarev:
@bborn Hey! If you're able to fix the specs by any chance -- I'm happy to merge! I just haven't had the time to look into it.

bborn:
@andreibondarev CI is passing

langchain.gemspec
```diff
@@ -55,7 +55,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "nokogiri", "~> 1.13"
   spec.add_development_dependency "open-weather-ruby-client", "~> 0.3.0"
   spec.add_development_dependency "pg", "~> 1.5"
-  spec.add_development_dependency "pgvector", "~> 0.2"
+  spec.add_development_dependency "pgvector", "~> 0.2.1"
```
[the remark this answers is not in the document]
andreibondarev: @technicalpickles Just out of curiosity -- what's the benefit of doing that?

andreibondarev:
@bborn Thank you for getting this over the finish line!

PR #179, opened 2023-06-14
Change llm prompt from json to yaml

mattlindsey:
Tested for openai llm.  Will test from google palm when that is working for me.


andreibondarev:
@mattlindsey This is perfect! Much more readable!

PR #181, opened 2023-06-14
Improve db schema dump

mattlindsey:
Improves the database schema dump to easy to digest format, which should help the llm, and it is smaller in size allowing for larger databases to be used.

Completes https://github.com/andreibondarev/langchainrb/issues/138

Adds parameters to include or exclude tables from the schema, which can be used especially when prompt context size is blowing up.
```ruby
db = Langchain::Tool::Database.new(connection_string: "xxx", tables: [:users])
db = Langchain::Tool::Database.new(connection_string: "xxx", exclude_tables: [:catalog_table1, :catalog_table2])

### CODE REVIEW

File: spec/langchain/agent/sql_query_agent/sql_query_agent_spec.rb [hunk 1]
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 RSpec.describe Langchain::Agent::SQLQueryAgent do
-  let(:db) { Langchain::Tool::Database.new(connection_string: "mock:///") }
+  let(:db) { Langchain::Tool::Database.new(connection_string: "mock://postgres") }
```
mattlindsey, line 4: Note that `mock:///` can't be used here, since that class doesn't implement all methods (like `table`).

mattlindsey:
This PR should solve #187 if they are able to pass in which tables to include and/or exclude to make the schema dump much smaller.  This is what other projects do.

mattlindsey:
@andreibondarev Could you take a look at this one when you get a chance?  The only quirk I saw in testing is that if you exclude a table it will still dump foreign keys on other tables that reference it.

andreibondarev:
> @mattlindsey: [That other] PR doesn't really solve much as the llm seems to understand even the ugly current dump.

This is true, but the updated DB schema dump is shorter => less tokens to be used.

I'll merge this PR for now but a few things that I would do:
1. Put the foreign keys under the `create table (...)` statement as indicated in that research paper.
2. Maybe rename `except_tables:` to `exclude_tables:`.

andreibondarev:
@mattlindsey Per usual, great work!

PR #182, opened 2023-06-15
Update limits for OpenAI token length validator

alchaplinsky:
## Changes
Add token validator limits for new `gpt-3.5-turbo-16k` models

https://platform.openai.com/docs/models/moderation


andreibondarev:
@alchaplinsky 🥇

PR #188, opened 2023-06-17
Simple text chunking

andreibondarev:
Simple text chunking using the [baran](https://github.com/moekidev/baran) gem.

Example usage:
```ruby
irb(main):001:0> Langchain::Loader.new("spec/fixtures/loaders/sample.docx").load.chunks
=> [{:text=>"Lorem ipsum \n\nLorem ipsum dolor sit amet, consectetur adipiscing elit. Nunc ac faucibus odio.", :cursor=>0}, {:text=>"Vestibulum neque massa, scelerisque sit amet ligula eu, congue molestie mi. Praesent ut varius sem. Nullam at porttitor arcu, nec lacinia nisi. Ut ac dolor vitae odio interdum condimentum. Vivamus dapibus sodales ex, vitae malesuada 
... }]
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,9 +1,4 @@
 PATH
-  remote: ../baran
-  specs:
-    baran (0.1.6)
-
-PATH
   remote: .
   specs:
     langchainrb (0.5.5)
@@ -38,6 +33,7 @@ GEM
     afm (0.2.2)
     ai21 (0.2.0)
     ast (2.4.2)
+    baran (0.1.6)
     builder (3.2.4)
     byebug (11.1.3)
     childprocess (4.1.0)
@@ -318,7 +314,6 @@ PLATFORMS
 
 DEPENDENCIES
   ai21 (~> 0.2.0)
-  baran!
   chroma-db (~> 0.3.0)
   cohere-ruby (~> 0.9.4)
   docx (~> 0.8.0)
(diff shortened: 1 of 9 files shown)


alchaplinsky (approved):
Nice 👍

andreibondarev:
@alchaplinsky We should figure out how the output of this `chunks` method plugs into the [add_data](https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/vectorsearch/base.rb#L148-L157) or add_texts when indexing data into vector search DBs

alchaplinsky:
@andreibondarev Do different vector search databases require different chunk size?

andreibondarev:
> @andreibondarev Do different vector search databases require different chunk size?

It's not at the vector search DB level, it's a concern for LLMs. 

For example when data is added to Chroma DB (and all other vector search DBs (almost) work the same way) [here](https://github.com/andreibondarev/langchainrb/blob/0.5.5/lib/langchain/vectorsearch/chroma.rb#L40-L43) the text itself is added along with the associated embedding.

If we generate 1 embedding from the whole large text and then that whole text gets retrieved and passed to the LLM to synthesize the answer. The prompt literally looks like this:
```
Context:
{context} # The whole text here

Question: {question}

Answer:
```

The whole text might exceed the LLM context window so it needs to be split into smaller chunks.

PR #190, opened 2023-06-17
Refactor/refactor openai compose chat messages

foreverLoveWisdom:
- Refactor compose_chat_messages to improve maintainability

### DIFF
--- a/lib/langchain/llm/openai.rb
+++ b/lib/langchain/llm/openai.rb
@@ -72,22 +72,12 @@ module Langchain::LLM
     def chat(prompt: "", messages: [], context: "", examples: [], **options)
       raise ArgumentError.new(":prompt or :messages argument is expected") if prompt.empty? && messages.empty?
 
-      parameters = compose_parameters DEFAULTS[:chat_completion_model_name], options
-      parameters[:messages] = compose_chat_messages(prompt: prompt, messages: messages, context: context, examples: examples)
-      parameters[:max_tokens] = validate_max_tokens(parameters[:messages], parameters[:model])
+      parameters = compose_chat_parameters(DEFAULTS[:chat_completion_model_name], prompt: prompt, messages: messages, context: context, examples: examples, **options)
 
-      if (streaming = block_given?)
-        parameters[:stream] = proc do |chunk, _bytesize|
-          yield chunk.dig("choices", 0, "delta", "content")
-        end
-      end
-
-      response = client.chat(parameters: parameters)
-
-      raise "Chat completion failed: #{response}" if !response.empty? && response.dig("error")
-
-      unless streaming
-        response.dig("choices", 0, "message", "content")
+      if block_given?
+        stream_response(parameters) { |chunk| yield chunk.dig("choices", 0, "delta", "content") }
+      else
+        complete_response(parameters)
       end
     end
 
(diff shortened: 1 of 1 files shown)


andreibondarev:
@foreverLoveWisdom Is the [other PR](https://github.com/andreibondarev/langchainrb/pull/189/files) just a duplicate?

lib/langchain/llm/openai.rb
```diff
@@ -119,24 +129,38 @@ def compose_parameters(model, params)
     def compose_chat_messages(prompt:, messages:, context:, examples:)
       history = []
 
-      history.concat transform_messages(examples) unless examples.empty?
+      add_examples_to_history(examples, history)
```
alchaplinsky, line 132: I'm not convinced slicing this method in these smaller `add_..._to_history` methods makes it more readable/maintainable. First of all, it is a **conditional** add whatever to history, and the method name does not hit there's a condition. So I would move these checks `unless examples.empty?` back to `compose_chat_messages` method.
And if you move those back the only "logic" that goes to those simple add methods is just `history.concat(...)` which is already pretty self-explanatory and there's not a ton of sense to create these tiny methods.

alchaplinsky (commented):
Overall, this refactoring is focused more on slicing `chat` and `compose_chat_messages ` methods into a number of smaller methods, which is something I personally like if it makes sense. However, it doesn't add a lot of benefits/features, just slices already written code into chunks that are not even re-used.

I think at this point we're more interested in the contribution that adds more features or improves existing ones rather than cosmetical code refactoring. Since this library's code changes often and dramatically it doesn't make a ton of sense at the moment. 

However, if you want to refactor some things within the existing code, I'd look at code duplication and the fact that we need a common way to compose params for multiple methods on this class and not only focus on `chat` method. Unifying that would actually bring benefit to developers.

foreverLoveWisdom:
> Overall, this refactoring is focused more on slicing `chat` and `compose_chat_messages ` methods into a number of smaller methods, which is something I personally like if it makes sense. However, it doesn't add a lot of benefits/features, just slices already written code into chunks that are not even re-used.
> 
- "However, it doesn't add a lot of benefits/features, just slices already written code into chunks that are not even re-used.' 
=> Thank you for the feedback 🙏 I agree and disagree at the same time. Agreeing because one of the main benefits of refactoring is for code-reuse. Nevertheless, code refactoring does more than that, for example, refactoring helps lead to code that is easy to read -> easy to maintain -> reducing the number of bugs. 
- Moreover, just because a method cannot be reused now, does not mean it can't be reused later. It makes the developers more conscious of reusing these methods in the future if the situation asks for them. 
- The last benefit, refactoring helps simplify things by introducing the right abstraction into it(which I did in this case) and hiding the details of implementations(and making the implementations easier to understand through the new abstractions).
- In short, my refactoring reduces its complexity(FLOG) before and after. But the FLOG is just one of the factors, the most important reason is to make the code easy to read and comprehend -> reduces potential bugs and improve maintainability

> Since this library's code changes often and dramatically it doesn't make a ton of sense at the moment.

=> I got it, and it still makes sense to refactor even if the codebase does frequently change (for the above reasons). A codebase that changes often still needs to be easy to read and well-structured for maintainability effort.

> However, if you want to refactor some things within the existing code, I'd look at code duplication and the fact that we need a common way to compose params for multiple methods on this class and not only focus on chat method. Unifying that would actually bring benefit to developers.

=> Extracting code duplication into a separate method does not help if the current method/class/codebase is poorly-structured (messy abstraction level, multiple if-else-unless checks here and there...)

=> So, Let me know what you think of my points, and then I will see how I can contribute to this project. 😄

lib/langchain/llm/openai.rb
foreverLoveWisdom: - I agree the `history.concat(...)` are self-explanatory but there are many of them, and they add noises to this private interface, the output of this interface is to compose the chat messages. The noise details should be just extracted and put in separate methods, later, if we want to expose this private interface outside, it will make it easier to maintain. 😄

foreverLoveWisdom:
> @foreverLoveWisdom Is the [other PR](https://github.com/andreibondarev/langchainrb/pull/189/files) just a duplicate?

- Actually, I was thinking of creating 2 PRs for 2 separate refactorings, but I accidentally branched out from the previous PR for this, so yeah, let me close the other one. Sorry for any inconvenience. 🙇

alchaplinsky:
@foreverLoveWisdom 
> Extracting code duplication into a separate method does not help if the current method/class/codebase is poorly-structured (messy abstraction level, multiple if-else-unless checks here and there...)

That's what I'm saying take a look at the entire class and make an improvement that would bring more benefit. Like, unifying params composition and calculating max_tokens that is duplicated in other methods. If you can extract that logic into nice re-usable methods that would be much more valuable refactoring.

andreibondarev:
I've completely refactored the `Langchain::LLM::OpenAI` class in an upcoming PR.

PR #191, opened 2023-06-17
Improve Conversation class to handle limits

alchaplinsky:
## Motivation
Langchain's `Conversation` class should be able to handle token limit exceeded situations and be smart enough to drop older messages and keep the conversation going even if it is too long for LLM to process all the messages.

## Changes
- Allow passing `messages` to `Conversation` initializer to assign some message history to the conversation instantly.
- Update `TokenLimitExceeded` to have a custom initializer. So that we can pass `token_overflow` parameter which can then be obtained from the exception.
- Rescue from `TokenLimitExceeded` exception and try removing some older messages to give LLM some space (20 tokens) for a response.
- Handle a case when there's a single message left but the length is too big for the model

## Ideas
Currently, we drop messages directly from `messages` array which is the entire conversation history. This leads to losing the entire conversation unless those messages are not stored somewhere by the code using Conversation.

**Idea worth exploring:** Maybe we should keep a separate copy of messages in a `history` array and use `messages` for communicating with LLM. This way we'll be dropping messages from `messages` and will keep the entire conversation in `history`. This will lead to a memory usage increase, though. So not doing it for now.


spec/langchain/conversation_spec.rb
```diff
@@ -115,5 +115,58 @@
...
+      let(:messages) { [] }
+
+      subject { described_class.new(llm: llm, messages: messages) }
+
+      before do
+        allow(llm).to receive(:client).and_return(client)
+        allow(client).to receive(:chat).and_return(response)
+      end
+
+      context "a single pormpt that exceeds the token limit" do
```
andreibondarev, line 131: Typo: prompt

andreibondarev (commented):
Looks good! 👍🏻

Have you tested it with Google Palm as well?

Also -- should we treat any of the system messages differently? Since those are the instructions for the LLM.

alchaplinsky:
@andreibondarev Good point about `system` message. Will add a test case and update the logic.

alchaplinsky:
Alright, I've added `context` and `examples` to the test. Also made this logic support for `GooglePalm` model and did some refactoring along the way.

andreibondarev (approved):

PR #192, opened 2023-06-17
Sequential agent that just runs tools in order

bborn:
I've found many use cases where having an agent that can just run tools in order (input => output => input etc.) is really useful.

Here's a first pass at that. I'd like to add state to this (so e.g. Tool 4 can access some variable that was set in Tool 1), but haven't been able to figure it out yet.

Sequential Agent use case example

Tool 1 => determine the language of the question
Tool 2 => translate the question to english
Tool 3 => lookup the answer to the question
Tool 4 => translate the answer back into original language

E.g.:
"Que es el color del cielo?"
"What is the color of the sky?"
"The color of the sky is blue"
"El color del cielo es azul."

I think this would require some way of maintaining state along the way, so each Tool can access e.g. a state hash (`state = {original_language: 'es'}`). Would love to hear some ideas on how to do it.

### DIFF
--- a/lib/langchain.rb
+++ b/lib/langchain.rb
@@ -76,6 +76,7 @@ module Langchain
     autoload :Base, "langchain/agent/base"
     autoload :ChainOfThoughtAgent, "langchain/agent/chain_of_thought_agent/chain_of_thought_agent.rb"
     autoload :SQLQueryAgent, "langchain/agent/sql_query_agent/sql_query_agent.rb"
+    autoload :SequentialAgent, "langchain/agent/sequential_agent/sequential_agent.rb"
   end
 
   module Tool
--- a/lib/langchain/agent/sequential_agent/sequential_agent.rb
+++ b/lib/langchain/agent/sequential_agent/sequential_agent.rb
@@ -1,0 +1,41 @@
+require "json"
+
+module Langchain::Agent
+  class SequentialAgent < Base
+    attr_accessor :tools
+
+    # Initializes the Agent
+    #
+    # @param llm [Object] The LLM client to use
+    # @param tools [Array] The tools to use
+    # @return [ChainOfThoughtAgent] The Agent::SequentialAgent instance
+    def initialize(llm:, tools:)
+      # NOTE: don't need to validate tools here. For sequential agent it's fine to have duplicate tools
+
+      @tools = tools || []
+    end
+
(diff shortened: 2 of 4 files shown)


andreibondarev:
@bborn Does this tool not use an LLM in any capacity?

mattlindsey:
> I think this would require some way of maintaining state along the way, so each Tool can access e.g. a state hash (`state = {original_language: 'es'}`).

Maybe a class variable `@@context` on Langchain::Tool?

bborn:
@andreibondarev no - the agent doesn't use the LLM. In my app (AgentHQ) I'm instantiating the agent class dynamically, so I needed the agent to receive the same arguments as all the other agent classes. Probably there's a better way around this.

@mattlindsey my main use case here involves using the RubyCodeInterpreter tool in sequence with other tools. So the challenge is, how do you pass the state in to the code interpreter, allow it to modify the state, and hand it back to the agent? Or in the case of other tools (e.g. Search) - how does the tool access the agent state and do something with it? For example, the Search tool might need to use the `original_language` from the first tool to only look for results in a certain language.

mattlindsey:
@bborn I guess I was thinking you could look at it as all tools sharing a `context` or `state` which the agent could also access through a `self.state` or `self.get_state` method on Langchain::Tool, but maybe that's not quite right.

bborn:
@mattlindsey  that's exactly right. I'm just not clear on the right way to do it :)

mattlindsey:
@bborn I'm certainly no expert, but seems a variable `@@state` on `Langchain::Tool::Base` with a `self.get_state` method might work.

andreibondarev:
@bborn Is [this](https://github.com/hwchase17/langchain/blob/9c0cb90997db9eb2e2a736df458d39fd7bec8ffb/langchain/chains/sequential.py) the equivalent that exists in the original Langchain? I'm trying to figure out why or why not should this workflow exist in Langchain.rb.

bborn:
@andreibondarev there isn't an equivalent Agent class. The SequentialChain is probably the most similar:

https://python.langchain.com/docs/modules/chains/foundational/sequential_chains

Except that it's a chain of prompts, not Tools.

And https://python.langchain.com/docs/modules/chains/foundational/sequential_chains#memory-in-sequential-chains is basically what I'm getting at with keeping state between tools.

PR #194, opened 2023-06-18
Improve OpenAI#chat documentation

technicalpickles:
I've started using this, and ended up having to refer to the source. This tries to describe the parameters a bit better, and include some examples to get started with.

### DIFF
--- a/lib/langchain/llm/openai.rb
+++ b/lib/langchain/llm/openai.rb
@@ -60,14 +60,37 @@ module Langchain::LLM
     end
 
     #
-    # Generate a chat completion for a given prompt
+    # Generate a chat completion for a given prompt or messages.
+    #
+    # == Examples
+    #
+    #     openai.chat prompt: "When was Ruby first released?"
+    #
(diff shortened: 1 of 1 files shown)


lib/langchain/llm/openai.rb
```diff
@@ -60,14 +60,37 @@ def complete(prompt:, **params)
...
+    #
+    # == Examples
+    #
+    #     openai.chat prompt: "When was Ruby first released?"
+    #
+    #     openai.chat context: "You are RubyGPT, a helpful chat bot for helping people learn Ruby", prompt: "Does Ruby have a REPL like IPython?"
+    #
+    #     openai.chat messages: [
+    #       {
+    #         role: "system",
```
alchaplinsky, line 73: You don't actually have to pass a `system` message. You can pass `context` instead and the client will prepend a system message to message history.

technicalpickles: Right, you don't have to, but this is also a valid way to chat. I can include more examples.

I think I would tend to prefer this way, since it directly maps to the OpenAI REST API: https://platform.openai.com/docs/api-reference/chat/create

PR #196, opened 2023-06-18
Allow passing default options when instantiating the llm

bborn:
This makes it possible to set the LLM model when instantiating the class, rather than just when running the completion method (e.g. `complete`). Helpful when running an Agent and you want it to use a particular model.

### DIFF
--- a/lib/langchain/llm/ai21.rb
+++ b/lib/langchain/llm/ai21.rb
@@ -32,7 +32,7 @@ module Langchain::LLM
     # @return [String] The completion
     #
     def complete(prompt:, **params)
-      parameters = complete_parameters @defaults[:model], params
+      parameters = complete_parameters params
 
       response = client.complete(prompt, parameters)
       response.dig(:completions, 0, :data, :text)
@@ -52,10 +52,8 @@ module Langchain::LLM
 
     private
 
-    def complete_parameters(model, params)
-      default_params = {model: model, temperature: @defaults[:temperature]}
-
-      default_params.merge(params)
+    def complete_parameters(params)
+      @defaults.dup.merge(params)
     end
   end
 end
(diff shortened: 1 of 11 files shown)


bborn:
@andreibondarev confused about the failing CI test. It says:

```
rspec ./spec/langchain/conversation_spec.rb:224 # Langchain::Conversation#message with length limit exceeded with PaLM2 LLM message history exceeds the token limit should drop 2 first messages and call an API
```

All tests are passing locally.

But I don't see a line #224 in that file at all:

https://github.com/bborn/langchainrb/blob/llm-default-options/spec/langchain/conversation_spec.rb

andreibondarev:
@alchaplinsky Are you able to take a glance at this?

lib/langchain/llm/ai21.rb
```diff
@@ -26,7 +32,9 @@ def initialize(api_key:)
     # @return [String] The completion
     #
     def complete(prompt:, **params)
-      response = client.complete(prompt, params)
+      parameters = complete_parameters @defaults[:model], params
```
alchaplinsky, line 35: Why do we pass `@defaults[:model]` here?

lib/langchain/llm/ai21.rb
```diff
@@ -41,5 +49,13 @@ def summarize(text:, **params)
       response = client.summarize(text, "TEXT", params)
       response.dig(:summary)
     end
+
+    private
+
+    def complete_parameters(model, params)
+      default_params = {model: model, temperature: @defaults[:temperature]}
```
alchaplinsky, line 56: since `model` is exactly `@defaults[:model]` can't we just do?
```ruby
def complete_parameters(params)
   @defaults.dup.merge(params)
end
```

alchaplinsky (approved):
Looks good 👍

andreibondarev:
@bborn Looks great, thank you!! 👏🏻

PR #200, opened 2023-06-20
add max token length validator for Cohere; add `truncate` to default_options

jinshen1983:
1. add max token length validator for Cohere
2. add `truncate` to default_options
> During the test, Cohere `tokenize` API  returns an input with token length 252. The model `command` max token length is 4096, so I set the `max_tokens` to 4096 - 252 = 3844. But the `generate`'s return has 3846 token length, which exceeds the max allowed token length and throws an exception. So, I add the default `truncate` option to `START`.

### DIFF
--- a/lib/langchain/utils/token_length/cohere_validator.rb
+++ b/lib/langchain/utils/token_length/cohere_validator.rb
@@ -8,7 +8,7 @@ module Langchain
       # It is used to validate the token length before the API call is made
       #
 
-      class CohereValidator
+      class CohereValidator < BaseValidator
         TOKEN_LIMITS = {
           # Source:
           # https://docs.cohere.com/docs/models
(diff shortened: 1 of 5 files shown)


lib/langchain/utils/token_length/cohere_validator.rb
```diff
@@ -0,0 +1,64 @@
...
+        }.freeze
+        #
+        # Calculate the `max_tokens:` parameter to be set by calculating the context length of the text minus the prompt length
+        #
+        # @param content [String | Array<String>] The text or array of texts to validate
+        # @param model_name [String] The model name to validate against
+        # @return [Integer] Whether the text is valid or not
+        # @raise [TokenLimitExceeded] If the text is too long
+        #
+        def self.validate_max_tokens!(content, model_name, client)
```
alchaplinsky, line 33: This method is already implemented in `BaseClass` so that we don't need to repeat it: https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/utils/token_length/base_validator.rb#L15

alchaplinsky:
hey @jinshen1983, thanks for putting this together. Could you also inherit from `BaseValidator` and re-use the logic implemented there.

andreibondarev:
@jinshen1983 I went ahead and bumped up the cohere gem for you with your fix: https://github.com/andreibondarev/langchainrb/pull/201

lib/langchain/utils/token_length/cohere_validator.rb
jinshen1983: looks like I missed the latest code in my local. let me merge the latest code.

andreibondarev (approved):
@jinshen1983 Thank you for the contribution, and great work! 🚢

Issue #177, opened 2023-06-13
[Proposal] Improvements to the ChainOfThoughtAgent


andreibondarev:
~~1. Rename `ChainOfThoughtAgent` to `ReActAgent` because that's what it actually is. This is what "chain of thought" is: https://learnprompting.org/docs/intermediate/chain_of_thought; and this is what "ReAct" actually is: https://arxiv.org/abs/2210.03629~~
2. Try asking the LLM for a JSON output format -- may or may not be more accurate. Source: 
<img width="1504" alt="image" src="https://github.com/andreibondarev/langchainrb/assets/541665/5eaf8063-d265-468f-9f4f-a448bd0f0df7">. i.e. instead of [RegEx-ing strings](https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/agent/chain_of_thought_agent/chain_of_thought_agent.rb#L70), we may be able to read off JSON keys instead.

(Separate PRs please).

mattlindsey:
Improvement 2 sounds nice since subsequent calls to the llm `complete` will add llm responses grouped in JSON clumps, which will be easier to understand.

mattlindsey:
How about I do 1 PR to rename the agents to ReAct and SQLQueryChain?

andreibondarev:
"SqlQueryAgent" doesn't work anymore? I'd like to refer or avoid the notion of "chains" for the time being.

mattlindsey:
My feeling is SQLQueryAgent really isn't semi-autonomous being limited to 2 steps and 1 tool, but that could change. So that's fine.

arthurhess:
Hi! I'm new to this project and I'd love to work on these. May I?

PR #207, opened 2023-06-21
Refactor: rename ChainOfThoughtAgent to ReActAgent

arthurhess:
Breaking change. Implementing the first proposal listed in https://github.com/andreibondarev/langchainrb/issues/177


arthurhess:
I'm assuming the changelog should be updated later, but I can draft something now, if it helps.

arthurhess:
I only force-pushed because I realized my previous commit was signed with the wrong key.

image

lib/langchain.rb
```diff
@@ -74,7 +74,7 @@ def logger=(logger)
 
   module Agent
     autoload :Base, "langchain/agent/base"
-    autoload :ChainOfThoughtAgent, "langchain/agent/chain_of_thought_agent/chain_of_thought_agent.rb"
+    autoload :ReActAgent, "langchain/agent/re_act_agent/re_act_agent.rb"
```
andreibondarev, line 77: Maybe the folder and filename could just be `langchain/agent/react_agent/react_agent.rb"`?

arthurhess: Right. I was trying to be consistent with the ReAct casing, but it does look weird. Thanks for pointing it out.

PR #211, opened 2023-06-22
🚃 `Langchain::ActiveRecord::Hooks` module for vectorsearch capabilities

andreibondarev:
Jun-22-2023 15-29-38

TODOs:
- [x] Unit testing
- [x] `similarity_search()` should return actual ActiveRecord objects
- [x] Add info to the README


lib/langchain/active_record/hooks.rb
```diff
@@ -0,0 +1,58 @@
...
+    # Delete the default schema to start over
+    #     Recipe.class_variable_get(:@@provider).client.schema.delete class_name: "Recipes"
+    #
+    module Hooks
+      def self.included(base)
+        base.extend ClassMethods
+      end
+
+      def index_to_vectorsearch
+        self.class.class_variable_get(:@@provider).add_texts(texts: self.to_json)
```
moekidev, line 39: `self.to_json` looks good. We can deal with any vector DB.

moekidev:
Nice. The use of LLM within the familiar MVC architecture would lower the psychological barrier for newcomers.

moekidev:
The provider is defined in the config, and only the index name is specified in the model.

andreibondarev:
> I assume that the provider is defined in the config, and only the index name is specified in the model.

Sorry -- I'm not sure what you mean?

moekidev:
```
Langchain::ActiveRecord.configure do |config|
  config.provider_name = :weaviate
  config.provider_url = 
  config.provider_api_key
end
```

```
class Product < ActiveRecord::Base
  include Langchain::ActiveRecord::Hooks

  vector_index_name "Recipes"
end
```

moekidev:
> I once worked on a Rails application that connected to 10+ different databases.

OK😂 I guess we should adopt a design that allows us to choose a vector DB for each model.

technicalpickles:
As this is right now, I don't think it adds a ton of value to for a user, as opposed to just hooking it into yourself. Some things I can think of that would make it more compelling:

- less code to set up a vector store in a model
- define the vectorstores more centrally
- handle things like index names automatically, ie use the model name

I have been reading up for activestorage recently, and I think it gives a pretty good pattern. You have a `config/storage.yml` (similar to `config/database.yml`) that defines all the different service you might use:

```yaml
local: # the key is how you use refer to it from the code
  service: Disk # this corresponds to a class on the backend
  root: <%= Rails.root.join("storage") %> # ERB is supported

test:
  service: Disk
  root: <%= Rails.root.join("tmp/storage") %>

amazon:
  service: S3
  access_key_id: ""
  secret_access_key: ""
  bucket: ""
  region: "" # e.g. 'us-east-1'
```

Models can also choose to use something different:

```ruby
class User < ApplicationRecord
  has_one_attached :avatar, service: :s3
end
```
For the other stuff, I'm kinda curious how much ends up being vectorstore specific 🤔  Like the index, at least the naming conventions, would be specific weaviate if I am reading some of the comments correctly?

andreibondarev:
> 1. less code to set up a vector store in a model
> 2. define the vectorstores more centrally
> 3. handle things like index names automatically, ie use the model name

@technicalpickles The PR you opened takes care of # 1.

2. I agree that some sort of YAML config would be nice but do you think we need it now?
3. I'll make `index_name:` default to the `.table_name`.

> For the other stuff, I'm kinda curious how much ends up being vectorstore specific
This is a very good question. When you look at the classes under `langchain/vectorsearch/*`, what does it look like to you? There's certainly difference, similar to how there's differences between the different LLMs, but I wonder if we can continue drawing out a common DSL for all of them with `opts={}` to go deeper with each of them if needed.

lib/langchain/active_record/hooks.rb
```diff
@@ -0,0 +1,56 @@
...
+    # Delete the default schema to start over
+    #     Recipe.class_variable_get(:@@provider).client.schema.delete class_name: "Recipes"
+    #
+    module Hooks
+      def self.included(base)
+        base.extend ClassMethods
+      end
+
+      def index_to_vectorsearch
+        self.class.class_variable_get(:@@provider).add_texts(texts: to_json)
```
technicalpickles, line 37: I was wondering how well this works as a default? It would end up using `as_json`, which defaults to `attributes` iirc, so all the columns for the record. That includes things like created/updated dates, id, etc. Is that useful to index?

Maybe could add an instance method like `as_vector`, which defaults to attributes. That way, a user can re-implement that for their models. [relevant rails docs](https://api.rubyonrails.org/classes/ActiveModel/Serializers/JSON.html)

lib/langchain/active_record/hooks.rb
```diff
@@ -0,0 +1,56 @@
...
+        def vectorsearch(provider:)
+          class_variable_set(:@@provider, provider)
+        end
+
+        def similarity_search(query, k: nil)
+          class_variable_get(:@@provider).similarity_search(
+            query: query,
+            k: k
+          )
+        end
```
technicalpickles, line 52: What does this end up returning? I think it'd be the document indexed, which I think would be the JSON? 🤔 

I would expect this to return ActiveRecord objects. If you can get the ids back from the similarity search, would be able to use `find(ids)` to get an array of records.

lib/langchain/vectorsearch/weaviate.rb
```diff
@@ -24,6 +24,8 @@ def initialize(url:, api_key:, index_name:, llm:)
         url: url,
         api_key: api_key
       )
+
+      # Weaviate requires the class name to be Capitalized: https://weaviate.io/developers/weaviate/configuration/schema-configuration#create-a-class
       @index_name = index_name
```
technicalpickles, line 29: Maybe should add some validations here? Raise if it's not capitalized.

lib/langchain/active_record/hooks.rb
```diff
@@ -0,0 +1,56 @@
...
+    #     class Product < ActiveRecord::Base
+    #       vectorsearch provider: Langchain::Vectorsearch::Weaviate.new(
+    #                    api_key: ENV["WEAVIATE_API_KEY"],
+    #                    url: ENV["WEAVIATE_URL"],
+    #                    index_name: "Recipes",
+    #                    llm: Langchain::LLM::OpenAI.new(api_key: ENV["OPENAI_API_KEY"])
+    #                 )
+    #
+    #       after_create :index_to_vectorsearch
+    #       after_update :index_to_vectorsearch
```
technicalpickles, line 21: `after_save` covers both of these, and is one less line.

```suggestion
    #       after_save :index_to_vectorsearch
```

lib/langchain/vectorsearch/weaviate.rb
andreibondarev: Basically Weaviate **will** automatically capitalize `"products"` to `"Products"` when you're creating an index but then will claim that it's not found if you search the `"products"` index later.

lib/langchain/active_record/hooks.rb
```diff
@@ -0,0 +1,89 @@
...
+    # * `vectorsearch` class method to set the provider
+    # * `similarity_search` class method to search for similar texts
+    # * `index_to_vectorsearch` instance method to index the text to the provider
+    #
+    # Usage:
+    #     class Recipe < ActiveRecord::Base
+    #       vectorsearch provider: Langchain::Vectorsearch::Weaviate.new(
+    #                    api_key: ENV["WEAVIATE_API_KEY"],
+    #                    url: ENV["WEAVIATE_URL"],
+    #                    index_name: "Recipes",
```
technicalpickles, line 16: I'd like to see this handled automatically. 

I was thinking this could be handled in a similar way to `ActiveRecord::Base.table_name`? 
 - https://api.rubyonrails.org/classes/ActiveRecord/ModelSchema/ClassMethods.html#method-i-table_name-3D
 - https://api.rubyonrails.org/classes/ActiveRecord/ModelSchema/ClassMethods.html#method-i-table_name

Giving it more thought though, the index name is needed at the time vector search is instantiated. Could make it work if we made it lazily instantiated. Maybe something like...

```ruby
vectorsearch weaviate: {
  url: ENV["WEAVIATE_URL"],
}

# then in the class method
def vectorsearch(providers = {})
  # only allow one to be passed in
  # providers.keys corresponds to the object
  provider = providers.keys.first
  self.class.class_variable_set(:@@provider_options, providers[provider))
  # ...
end

def vectorsearch_provider
  @vectorsearch_provider ||=  ... # lookup provider by name, pass in the options
end
```

lib/langchain/vectorsearch/weaviate.rb
technicalpickles: Right. I'm suggesting to either automatically capitalize, or raise an error if it's not capitalized already.

andreibondarev: I'll go with auto-capitalize then.

PR #215, opened 2023-06-24
Add distance parameter to weaviate similarity_search

mattlindsey:
image

Is that error due to `vectorizer: "none"` below?
```ruby
  # Create default schema
    def create_default_schema
      client.schema.create(
        class_name: index_name,
        vectorizer: "none",
        properties: [
          # TODO: Allow passing in your own IDs
          {
            dataType: ["text"],
            name: "content"
          }
        ]
      )
    end

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -17,7 +17,9 @@ weaviate = Langchain::Vectorsearch::Weaviate.new(
   url: ENV["WEAVIATE_URL"],
   api_key: ENV["WEAVIATE_API_KEY"],
   index_name: "",
-  llm: Langchain::LLM::OpenAI.new(api_key: ENV["OPENAI_API_KEY"])
+  llm: Langchain::LLM::OpenAI.new(api_key: ENV["OPENAI_API_KEY"]),
+  model_service: :openai,
+  model_api_key: ENV["OPENAI_API_KEY"]
 )
 
 qdrant = Langchain::Vectorsearch::Qdrant.new(
--- a/lib/langchain/llm/google_palm.rb
+++ b/lib/langchain/llm/google_palm.rb
@@ -16,7 +16,7 @@ module Langchain::LLM
       dimension: 768, # This is what the `embedding-gecko-001` model generates
       completion_model_name: "text-bison-001",
       chat_completion_model_name: "chat-bison-001",
-      embeddings_model_name: "embedding-gecko-001"
+      embeddings_model_name: "textembedding-gecko" # "embedding-gecko-001"
     }.freeze
     LENGTH_VALIDATOR = Langchain::Utils::TokenLength::GooglePalmValidator
 
--- a/lib/langchain/vectorsearch/weaviate.rb
+++ b/lib/langchain/vectorsearch/weaviate.rb
@@ -8,7 +8,7 @@ module Langchain::Vectorsearch
     # Gem requirements: gem "weaviate-ruby", "~> 0.8.0"
     #
     # Usage:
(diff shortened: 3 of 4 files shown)


mattlindsey:
Looks like Weaviate wants you to use their 'modules' to do `nearText`.  I see the following available in their docs:
image

mattlindsey:
@andreibondarev I think we would need to enable 'modules' for the different llms in your weaviate-ruby gem, if I'm reading this correctly (see https://weaviate.io/developers/weaviate/configuration/modules).  There's an environment variable ENABLE_MODULES to set and some code to add in that gem to determine the vectorization method to use depending on the llm passed into `Weaviate::Client.new`.

If that's true, we should work on the weaviate-ruby gem before finishing this PR.  Correct?

mattlindsey:
image

mattlindsey:
@andreibondarev I'm getting an authenication error.  Can you look at this if possible?  

New usage, which can hopefully be simplifed:

```ruby
 weaviate = Langchain::Vectorsearch::Weaviate.new(
  url: ENV["WEAVIATE_URL"],
  index_name: "Recipes",
  llm: Langchain::LLM::GooglePalm.new(api_key: ENV["GOOGLE_PALM_API_KEY"]), 
  model_service: :google_palm, 
  model_api_key: ENV["GOOGLE_PALM_API_KEY"], 
  project_id: "websiteone-153210" ). # Project ID in google cloud under which api key was created
  
  weaviate.create_default_schema_palm

  weaviate.add_texts ...

  weaviate.simularity_search ...
  ```
  
  Error:
  ```
  /Users/mattlindsey/.rvm/gems/ruby-3.2.1/gems/weaviate-ruby-0.8.3/lib/weaviate/query.rb:41:in `rescue in get': {"Recipes"=>["explorer: get class: vectorize params: vectorize params: vectorize params: vectorize keywords: remote client vectorize: connection to Google PaLM failed with status: 401 error: Request had invalid authentication credentials. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project."]} (Weaviate::Error)
	from /Users/mattlindsey/.rvm/gems/ruby-3.2.1/gems/weaviate-ruby-0.8.3/lib/weaviate/query.rb:5:in `get'
	from /Users/mattlindsey/github/langchainrb/lib/langchain/vectorsearch/weaviate.rb:135:in `similarity_search'

andreibondarev:
@mattlindsey The way Langchain.rb handles Weaviate is that it's not using any of it's modules because, for example, with vectorization we do the vectorization ourselves with any LLM passed in and send the ready vectors.

mattlindsey:
@andreibondarev I guess that's fine, but it looks like you can't do many types of queries easily without the modules and there are features like batching for performance.  In that case, I'll look into other ways to limit a query by distance.

andreibondarev:
@mattlindsey Sorry, I'll review this PR in a bit. But do you need to enable modules to pass the distance param?

mattlindsey:
@andreibondarev No problem and no hurry.  I don't actually have a use for it at the moment, but all of the ways to limit a query (https://weaviate.io/developers/weaviate/search/similarity#limit-the-results) seem to require nearText.  I'm guessing it's hard or inefficient to do with just the vectors.   We can leave this for later if you want...

mattlindsey:
This can actually be done pretty easily without modules, so I'll close this PR and submit another:
```ruby
    def similarity_search_by_vector(embedding:, k: 4, distance: 2.0)
      near_vector = "{ vector: #{embedding} distance: #{distance} }"

PR #217, opened 2023-06-24
Add default value to ids arg of Weaviate#add_texts

moekidev:


lib/langchain/vectorsearch/weaviate.rb
```diff
@@ -35,7 +35,7 @@ def initialize(url:, api_key:, index_name:, llm:)
     # Add a list of texts to the index
     # @param texts [Array] The list of texts to add
     # @return [Hash] The response from the server
-    def add_texts(texts:, ids:)
+    def add_texts(texts:, ids: [])
```
andreibondarev, line 38: The ids here intended to set the custom __id. Because apparently you can't set a proper Weaviate ID unless it's a uuid.

andreibondarev: I'm just wondering what kind of changes we'd need to make to update_texts()?

moekidev: Maybe we should have a dedicated method for ActiveRecord Hook?

PR #218, opened 2023-06-25
ActiveRecord Hooks with Pinecone

moekidev:


spec/langchain/vectorsearch/pinecone_spec.rb
```diff
@@ -133,6 +133,53 @@
...
+          metadata: {content: text},
+          values: embedding
+        }
+      ]
+    end
+
+    before do
+      vector = double(Pinecone::Vector)
+      allow(subject.llm).to receive(:embed).with(text: text).and_return(embedding)
+      allow(subject.client).to receive(:index).with(index_name).and_return(vector)
```
moekidev, line 174: `Pinecone::Client#index` returns `Pinecone::Vector` instance

https://github.com/ScotterC/pinecone/blob/99167444d7d16fe640f59afd81d1954b9b4a1a75/lib/pinecone/client.rb#L42-L44

andreibondarev (approved):
@moekidev Looks good!

lib/langchain/vectorsearch/pinecone.rb
```diff
@@ -51,6 +51,24 @@ def add_texts(texts:, namespace: "", metadata: nil)
...
 
+    # Update a list of texts in the index
+    # @param texts [Array] The list of texts to update
+    # @param ids [Array] The list of IDs to update
+    # @param namespace [String] The namespace to update the texts in
+    # @param metadata [Hash] The metadata to use for the texts
+    # @return [Array] The response from the server
+    def update_texts(texts:, ids:, namespace: "", metadata: nil)
+      texts.map.with_index do |text, i|
+        # Pinecone::Vector#update ignore args when it is empty
```
onerinas, line 62: `metadata` is `nil` here and pinecone checks using `.empty?` and hence results in
``` undefined method `empty?' for nil:NilClass```


In the active record hooks, update_texts is called without any metadata resulting in `nil` always and hence results in error always.

I'm not familiar with what metadata here is so not attempting to make a fix yet. 

Reached this pull request by following the code and hence the comment here :)

In `#add_texts`, we have `metadata` as `{content: text}` and here we

andreibondarev: @onerinas I'm wondering if we can just set it to `metadata: {}` and whether Pinecone would throw an error or not. 

Are you currently using this integration?

onerinas: Yes, I’m using the Pinecone integration with ActiveRecord. 

I have made few changes in this branch to suit my need but it’s in a rough state: https://github.com/onerinas/langchainrb/pull/1/files 

I’ll try to open a PR here once I’m in a good shape.

PR #219, opened 2023-06-25
ActiveRecord Hooks with Chroma

moekidev:
ref: https://github.com/mariochavez/chroma/blob/446f53d98a7118d9b676066e339e7d6a71b370bc/lib/chroma/resources/collection.rb#L148-L170


andreibondarev (approved):
💯 @moekidev

PR #220, opened 2023-06-25
feat: add Langchain::OutputParsers::OutputFixingParser to wrap a Langchain::OutputParser and handle invalid response

joshweir:
feat: add Langchain::OutputParsers::OutputFixingParser to wrap a Langchain::OutputParser and handle invalid response

* similar approach to langchainjs OutputFixingParser (https://js.langchain.com/docs/modules/prompts/output_parsers/#output-fixing-parser)
* OutputFixingParser wraps a Langchain::OutputParsers parser and calls LLM to fix the error


lib/langchain/output_parsers/fix.rb
```diff
@@ -0,0 +1,101 @@
...
+    #
+    # @return [Object] A new instance of the class
+    #
+    def self.from_llm(llm:, parser:, prompt: nil)
+      new(llm: llm, parser: parser, prompt: prompt || naive_fix_prompt)
+    end
+
+    private
+
+    private_class_method def self.naive_fix_prompt
```
andreibondarev, line 78: Should we put this into a `prompts/naive_fix_prompt.yaml` and read the template out of a file? Seems cleaner.

lib/langchain.rb
```diff
@@ -152,6 +152,7 @@ module ActiveRecord
   module OutputParsers
```
andreibondarev, line 152: Any chance that you could start a section in the README and document the Output Parsers, usage, etc.?

andreibondarev:
@joshweir Thank you for putting this together!

What I'm wondering is this: if we take this idea of having an LLM attempt to fix it's completion and forget for a second about how the Python/JS Langchains implement it -- what would "the Ruby" way for doing this look like?

Maybe something like the following?
```ruby
parser = Langchain::OutputParsers::StructuredOutputParser.from_json_schema(json_schema)

llm.complete(prompt: "...", parser: parser, naive_fix: true)
```

I'm just brain-storming... No need to change anything yet.

lib/langchain/output_parsers/fix.rb
joshweir: Sure, i'll fix it

lib/langchain.rb
joshweir: It'll be a pleasure

joshweir:
> @joshweir Thank you for putting this together!
> 
> What I'm wondering is this: if we take this idea of having an LLM attempt to fix it's completion and forget for a second about how the Python/JS Langchains implement it -- what would "the Ruby" way for doing this look like?
> 
> Maybe something like the following?
> 
> ```ruby
> parser = Langchain::OutputParsers::StructuredOutputParser.from_json_schema(json_schema)
> 
> llm.complete(prompt: "...", parser: parser, naive_fix: true)
> ```
> 
> I'm just brain-storming... No need to change anything yet.

I'm not too sure what the right `ruby` approach is here, I think your suggestion is good, question on your proposed example call:

```
llm.complete(prompt: "...", parser: parser, naive_fix: true)
```

If `!native_fix` then a fix prompt is required, perhaps we include optional `fix_prompt` which falls back to naive fix prompt? Or if we include `native_fix` boolean input, we need to validate that if its falsy then `fix_prompt` is required

andreibondarev:
Or you could just pass `naive_fix: true/false/"my_custom_prompt_string"` so it's the completion either doesn't get auto fixed when `false` or it does with your current default prompt when it's `true` or the developer supplies their own prompt.

Initially I'm not even sure I would bother with letting the developer pass their own prompt, it's just uses your default prompt.

lib/langchain.rb
joshweir: Added first draft of README section, if you could advise any feedback? cheers.

andreibondarev:
@joshweir I'm going to merge this, great job, and thank you for this feature!

Would you still be able to implement the interface we had talked about?
```ruby
llm.complete(prompt: "...", parser: parser, naive_fix: true|false|[prompt_string])
```

It's kind of a syntactic sugar ^^

PR #231, opened 2023-06-28
Introduce ConversationMemory

alchaplinsky:
## Motivation
ConversationMemory should be responsible for managing context/examples/messages. And conversation should be a simple  interface for sending messages to llm. 

## Chages
- Introduce `ConversationMemory` class to manage conversation history, context and examples
- Add `memory_strategy` option to `Conversation` to switch from truncating older messages to summarization


lib/langchain/conversation.rb
```diff
@@ -31,67 +28,58 @@ def initialize(llm:, **options, &block)
       @llm = llm
       @context = nil
       @examples = []
-      @messages = options.delete(:messages) || []
       @options = options
+      @memory = ConversationMemory.new(
+        llm: llm,
+        messages: options.delete(:messages) || [],
+        model: options[:model]
```
andreibondarev, line 35: Does the `model:` need to be passed explicitly or do we just look at what the `llm` object to see what the `llm.defaults` are?

andreibondarev (approved):

PR #237, opened 2023-07-01
support for OpenAI functions

michelson:
Hi there! Thanks for the library. I've been digging. Really good work!  
Questions, Are there the function_calls for open AI on the roadmap? I have integrated the latest version of open ai on my fork, and it kinda works, but there are some nuances:
+ To get the function calls is needed to get the whole response, not only the content of the choices
+ For that to work, I've added some options to get the whole chunk but
+ I had to remove the max_token option not sure why, but when the function's parameter is passed, it returns an error on max tokens, so due to that modification, specs fail.

Still figuring out the max token error.

### Example to run this on `Conversation` would be:

```ruby

require "langchain"
require "dotenv/load"

functions = [
    {
      name: "get_current_weather",
      description: "Get the current weather in a given location",
      parameters: {
        type: :object,
        properties: {
          location: {
            type: :string,
            description: "The city and state, e.g. San Francisco, CA",
          },
          unit: {
            type: "string",
            enum: %w[celsius fahrenheit],
          },
        },
        required: ["location"],
      },
    },
  ]


openai = Langchain::LLM::OpenAI.new( api_key: ENV["OPENAI_API_KEY"],  default_options: { chat_completion_model_name: "gpt-3.5-turbo-16k" }
)

chat = Langchain::Conversation.new(llm: openai, functions: functions, complete_chunk: true ) 

chat.set_context("You are the climate bot")

user_message = "what's the weather in NYC?"

puts chat.message(user_message)

```

> {"id"=>"chatcmpl-7XNFZLxn6hpisA0LHJ58ssQA4TERE", "object"=>"chat.completion", "created"=>1688188553, "model"=>"gpt-3.5-turbo-16k-0613", "choices"=>[{"index"=>0, "message"=>{"role"=>"assistant", "content"=>nil, "function_call"=>{"name"=>"get_current_weather", "arguments"=>"{\n  \"location\": \"New York, NY\"\n}"}}, "finish_reason"=>"function_call"}], "usage"=>{"prompt_tokens"=>90, "completion_tokens"=>19, "total_tokens"=>109}}

### example to run this on Vector ask:

```ruby
require "langchain"
require "dotenv/load"

functions = [
    {
      name: "get_current_weather",
      description: "Get the current weather in a given location",
      parameters: {
        type: :object,
        properties: {
          location: {
            type: :string,
            description: "The city and state, e.g. San Francisco, CA",
          },
          unit: {
            type: "string",
            enum: %w[celsius fahrenheit],
          },
        },
        required: ["location"],
      },
    },
  ]


openai = Langchain::LLM::OpenAI.new(
  api_key: ENV["OPENAI_API_KEY"],
  default_options: {
    chat_completion_model_name: "gpt-3.5-turbo-16k"
  }
)

client = Langchain::Vectorsearch::Qdrant.new(
  url: ENV["QDRANT_URL"],
  api_key: ENV["QDRANT_API_KEY"],
  index_name: ENV["QDRANT_INDEX"],
  llm: openai,
  functions: functions, 
  complete_chunk: true
)

chat = client.ask(question: "meaning of life or wheather on finland?" )

puts chat

## Response

```

> {"id"=>"chatcmpl-7XNdSA5hvs9pgaw6hcoXiQYztN5gz", "object"=>"chat.completion", "created"=>1688190034, "model"=>"gpt-3.5-turbo-16k-0613", "choices"=>[{"index"=>0, "message"=>{"role"=>"assistant", "content"=>"I'm sorry, I cannot provide the meaning of life. However, I can help you with the weather in Finland. Please wait a moment while I fetch the information.", "function_call"=>{"name"=>"get_current_weather", "arguments"=>"{\n  \"location\": \"Finland\"\n}"}}, "finish_reason"=>"function_call"}], "usage"=>{"prompt_tokens"=>2237, "completion_tokens"=>52, "total_tokens"=>2289}}

note the `complete_chunk: true` option, that will return the whole chunk instead the content


michelson:
this is wierd, when I pass functions with the max_tokens:

```curl
curl https://api.openai.com/v1/chat/completions -u :$OPENAI_API_KEY -H 'Content-Type: application/json' -d '{
  "model": "gpt-3.5-turbo-16k",
  "messages": [
    {"role": "assistant", "content": "would you like to schedule a meeting with some real agent?"},
    {"role": "user", "content": "not yet, thanks!"}
  ],
  "functions": [
    {
      "name": "generate_meeting_url",
      "description": "Get the intent of scheduling a meeting in the question",
      "parameters": {
        "type": "object",
        "properties": {
          "schedule": {
            "type": "boolean",
            "description": "intent of schedule, e.g. receive an afirmative response for a meeting invitation"
          },
          "unit": {
            "type": "string",
            "enum": ["url"]
          }
        },
        "required": ["schedule"]
      }
    }
  ],
  "max_tokens": 16356
}'
```
Reponse:

> This model's maximum context length is 16385 tokens. However, you requested 16452 tokens (27 in the messages, 69 in the functions, and 16356 in the completion). Please reduce the length of the messages, functions, or completion.

if the max_token option is removed then the response is successful

michelson:
I had added a skip condition for the max token when the functions is present

andreibondarev:
> @michelson if the max_token option is removed then the response is successful

But is the response complete or partially cut off?

michelson:
> > @michelson if the max_token option is removed then the response is successful
> 
> But is the response complete or partially cut off?

complete response, example:
```json
{
  "id": "chatcmpl-7XUdkD9XF1ySdAD7zyW4ixmsBOMKc",
  "object": "chat.completion",
  "created": 1688216960,
  "model": "gpt-3.5-turbo-16k-0613",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "You're welcome! If you have any other questions, feel free to ask."
      },
      "finish_reason": "stop"
    }
  ],
  "usage": {
    "prompt_tokens": 96,
    "completion_tokens": 17,
    "total_tokens": 113
  }
}
```

also, please try out the curl request that I've added here, the one I've pasted will fail displaying an error token. Also please try it by removing the max_token param to get the success response.

let me know if you have any questions, thanks!

michelson:
specs should be fixed by now @andreibondarev

andreibondarev:
@michelson It seems like we're not calculating the tokens correctly? Because the `functions:` are not even passed to the token validator.

michelson:
> @michelson It seems like we're not calculating the tokens correctly? Because the `functions:` are not even passed to the token validator.

I think that the issue is from the open ai itself, as you can see by trying the curl snippet that I've pasted in https://github.com/andreibondarev/langchainrb/pull/237#issuecomment-1615900030 
if you remove the max_token it pass, but if you keep the max_token it fails, also it is weird because the error is something like:

`    "message": "This model's maximum context length is 16385 tokens. However, you requested 16452 tokens (27 in the messages, 69 in the functions, and 16356 in the completion). Please reduce the length of the messages, functions, or completion.",
`

lib/langchain/llm/openai.rb
```diff
@@ -116,19 +116,20 @@ def chat(prompt: "", messages: [], context: "", examples: [], **options)
       parameters = compose_parameters @defaults[:chat_completion_model_name], options
```
andreibondarev, line 116: What if we actually exposed another method, like `chat_with_functions` just for OpenAI?

lib/langchain/vectorsearch/qdrant.rb
```diff
@@ -123,7 +124,7 @@ def ask(question:)
 
       prompt = generate_prompt(question: question, context: context)
 
-      llm.chat(prompt: prompt)
+      llm.chat(prompt: prompt, **@options)
```
andreibondarev, line 127: Instead of passing these options via the `Qdrant#ask` method I would probably recommend actually moving all of this method's logic into your own application:

```ruby
      search_results = similarity_search(query: question)

      context = search_results.map do |result|
        result.dig("payload").to_s
      end
      context = context.join("\n---\n")

      prompt = generate_prompt(question: question, context: context)

      llm.chat_with_functions(prompt: prompt, functions: functions)
```

lib/langchain/llm/openai.rb
michelson: Not sure about this, as this is an actual option of the chat prompt. why would you separate it on a specific function? 
My reasoning is that probably Open Ai will add more options and this library should be able to scale with that, without the need for more specific options. Also, I like the unified approach, adding another function for one API will kinda diverge the main idea.

lib/langchain/vectorsearch/qdrant.rb
michelson: This works like the Langchain::Conversation.new, in the way options can be passed through the LLM from that interface. It is handy as it seems the only way to pass options to the LLM from the vector search, Not only functions but other options too. Why would VectorSearch be different from Conversation in terms of options customization?

andreibondarev: Okay -- what's your thought on passing the `functions: []` to the `Langchain::Conversation` instance? E.g.:
```ruby
chat = Langchain::Conversation.new(llm: openai) 
chat.set_context("You are the climate bot")
chat.functions = functions
```

michelson: Definetely adding it as an accessor should work too. But given the `Langchain::Conversation` initializer has an `options` attribute already there I don't see any reason to not use that. You have the last word on this of course. 
Also, I would suggest that as the Conversation has the `options` that can be passed through the LLM it would make sense that the VectorSearch interface also has `options`. That way the API will be unified and will act as a transparent connector without needing to write/port logic to the apps.

andreibondarev: Would you please add the "chat.functions = functions" to the Conversation class? I don't think vector search classes should know about functions, doesn't seem like it's related.

michelson: I can add an accessor to the Conversation class, I expect to work on this tomorrow. Just I'm wondering why this does not fit your design needs.

```ruby
chat = Langchain::Conversation.new(llm: openai, functions: functions, complete_chunk: true ) 
```
It actually works as is without adding much code. 

note that we also need the complete_chunk option ... and eventually other options will be added from the open AI. 


Also this:
> I don't think vector search classes should know about functions, it doesn't seem like it's related.

Do you mean that Vector Search should not have support for GPT functions? or function options should not be passed through the Vector Search as options? 
I ask this as I Indeed have some use cases for this. VectorSearch + Function calls.

andreibondarev: I don't think function options should be passed through the vector search as options.

michelson: Examples:

1. conversation func calls https://gist.github.com/michelson/9c6ca9f10c39d81af0cc9ab68ec6adc7
2. vector search func calls: https://gist.github.com/michelson/bea785ddb0b3fa4faf9654cbe2373d73

Conversation essential bits are:

```ruby
chat = Langchain::Conversation.new(llm: openai)
chat.set_context("You are the climate bot")
chat.set_functions(functions)
```

Vector search important part is:

```ruby
client.llm.functions = functions
client.llm.complete_chunk = true
chat = client.ask(question: "create a users_controller")
```


Note the complete chunks in the conversation version will be added automatically on the `set_functions.` On the other hand the vector search version, you will need to add this option explicitly.

lib/langchain/conversation.rb
```diff
@@ -37,6 +37,11 @@ def initialize(llm:, **options, &block)
       @block = block
     end
 
+    def set_functions(functions)
+      @llm.functions = functions
+      @llm.complete_chunk = true
```
andreibondarev, line 42: Can we rename this variable to something else?

"chunk" refers to something very specific in the LLM world.

Perhaps "complete_response"?

lib/langchain/llm/openai.rb
```diff
@@ -116,19 +118,20 @@ def chat(prompt: "", messages: [], context: "", examples: [], **options)
...
       parameters[:messages] = compose_chat_messages(prompt: prompt, messages: messages, context: context, examples: examples)
       parameters[:max_tokens] = validate_max_tokens(parameters[:messages], parameters[:model])
-
+      parameters[:functions] = @functions unless @functions.nil?
```
andreibondarev, line 121: Do you still want to send an empty array?
```ruby
[].nil?
=> false
```

andreibondarev: Also whenever you define `attr_accessor ...`, I would be using the actual getter methods for those vars.

lib/langchain/llm/openai.rb
```diff
@@ -116,19 +118,20 @@ def chat(prompt: "", messages: [], context: "", examples: [], **options)
...
       if (streaming = block_given?)
         parameters[:stream] = proc do |chunk, _bytesize|
-          yield chunk.dig("choices", 0, "delta", "content")
+          yield chunk if @complete_chunk
+          yield chunk.dig("choices", 0, "delta", "content") unless @complete_chunk
         end
       end
 
+      parameters.delete(:max_tokens) if @functions
```
andreibondarev, line 129: Instead of deleting it here, can we just never set it in the first place? A few lines above:
```ruby
parameters[:max_tokens] = validate_max_tokens(parameters[:messages], parameters[:model]) unless functions
```

lib/langchain/vectorsearch/qdrant.rb
```diff
@@ -16,7 +16,7 @@ class Qdrant < Base
     # @param api_key [String] The API key to use
     # @param index_name [String] The name of the index to use
     # @param llm [Object] The LLM client to use
-    def initialize(url:, api_key:, index_name:, llm:)
+    def initialize(url:, api_key:, index_name:, llm:, **options)
```
andreibondarev, line 19: I think we said we'll be getting rid of the changes in this file, right?

michelson:
Hi @andreibondarev, I have updated the code, also I could add the examples and some doc notes

andreibondarev:
@michelson yes please! We need specs as well.

michelson:
> @michelson yes please! We need specs as well.

done!

examples/open_ai_qdrant_function_calls.rb
```diff
@@ -0,0 +1,44 @@
...
+  default_options: {
+    chat_completion_model_name: "gpt-3.5-turbo-16k"
+  }
+)
+
+client = Langchain::Vectorsearch::Qdrant.new(
+  url: ENV["QDRANT_URL"],
+  api_key: ENV["QDRANT_API_KEY"],
+  index_name: ENV["QDRANT_INDEX"],
+  # environment: ENV['PINECONE_ENVIRONMENT'],
```
michelson, line 36:
```suggestion
```

michelson:
Hi @andreibondarev do you expect to merge this soon or there are any comments or suggestions?

examples/open_ai_function_calls.rb
```diff
@@ -0,0 +1,41 @@
+require "langchain"
```
andreibondarev, line 1: @michelson Btw -- I ran this example and the other one and they both threw undefined errors method. Would you please double-check?

lib/langchain/llm/openai.rb
```diff
@@ -115,20 +117,27 @@ def chat(prompt: "", messages: [], context: "", examples: [], **options)
...
       parameters[:messages] = compose_chat_messages(prompt: prompt, messages: messages, context: context, examples: examples)
-      parameters[:max_tokens] = validate_max_tokens(parameters[:messages], parameters[:model])
+
+      unless functions
+        parameters[:max_tokens] = validate_max_tokens(parameters[:messages], parameters[:model])
+      end
+
+      unless functions.nil? || functions.empty?
+        parameters[:functions] = functions
+      end
```
andreibondarev, line 127: I think this can be simplified with:
```ruby
if functions
  parameters[:functions] = functions
else
  parameters[:max_tokens] = validate_max_tokens(parameters[:messages], parameters[:model])
end
```

lib/langchain/llm/openai.rb
```diff
@@ -115,20 +117,27 @@ def chat(prompt: "", messages: [], context: "", examples: [], **options)
...
 
       if (streaming = block_given?)
         parameters[:stream] = proc do |chunk, _bytesize|
-          yield chunk.dig("choices", 0, "delta", "content")
+          yield chunk if complete_response
+          yield chunk.dig("choices", 0, "delta", "content") unless complete_response
```
andreibondarev, line 132: Please change to `... if !complete_response` for consistency with below.

examples/open_ai_function_calls.rb
michelson: it works on my end, and it works on CI too it seems

lib/langchain/llm/openai.rb
michelson, line 132:
```suggestion
          yield chunk.dig("choices", 0, "delta", "content") if !complete_response
```

examples/open_ai_function_calls.rb
andreibondarev: The CI doesn't run the examples. They're meant to be run locally with "$ ruby examples/open_ai_function_calls.rb"

michelson: ok, and yes, examples works, do you have the qdrant configurated to some index?

lib/langchain/llm/openai.rb
michelson, line 127: if functions 
  parameters[:functions] = functions 
else parameters[:max_tokens] = 
  validate_max_tokens(parameters[:messages], parameters[:model]) 
end

michelson, line 127:
```suggestion
if functions 
  parameters[:functions] = functions 
else 
  parameters[:max_tokens] = validate_max_tokens(parameters[:messages], parameters[:model]) 
end
```

examples/open_ai_function_calls.rb
michelson: I do `bundle exec ruby  ./examples/open_ai_function_calls.rb` and `bundle exec ruby  ./examples/open_ai_qdrant_function_calls.rb` both work

andreibondarev:
@michelson Thank you for your patience and contribution here!

michelson:
> @michelson Thank you for your patience and contribution here!

Anytime. Thank you for bringing Langchain to Ruby

PR #240, opened 2023-07-01
Raise an error when token calculation API call fails

alchaplinsky:
Should make it clear that the API call failed instead of getting an error like https://github.com/andreibondarev/langchainrb/issues/238

### DIFF
--- a/lib/langchain/utils/token_length/google_palm_validator.rb
+++ b/lib/langchain/utils/token_length/google_palm_validator.rb
@@ -37,6 +37,9 @@ module Langchain
         #
         def self.token_length(text, model_name = "chat-bison-001", options)
           response = options[:llm].client.count_message_tokens(model: model_name, prompt: text)
+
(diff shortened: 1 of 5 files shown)


lib/langchain/utils/token_length/google_palm_validator.rb
```diff
@@ -37,6 +37,9 @@ class GooglePalmValidator < BaseValidator
         #
         def self.token_length(text, model_name = "chat-bison-001", options)
           response = options[:llm].client.count_message_tokens(model: model_name, prompt: text)
+
+          raise Langchain::LLM::ApiError.new(response["error"]["message"]) unless response["error"].nil?
```
andreibondarev, line 41: Do you want to extend this out to [here](https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/llm/openai.rb#L128) as well?

PR #248, opened 2023-07-11
PGVector#update_texts

dcluna:
This commit implements (tentative) pgvector support for upsert functionality, which as far as I can understand is all that is required to make it work correctly with the Rails integration.

This is open to discussion and modification - please let me know if there's anything to improve here.

### DIFF
--- a/spec/langchain/vectorsearch/pgvector_spec.rb
+++ b/spec/langchain/vectorsearch/pgvector_spec.rb
@@ -47,7 +47,7 @@ if ENV["POSTGRES_URL"]
       it "adds texts" do
         command = subject.add_texts(texts: ["Hello World", "Hello World"])
         expect(command).to be_a(::PG::Result)
-        expect(command.result_status).to eq(::PG::PGRES_COMMAND_OK)
+        expect(command.result_status).to eq(::PG::PGRES_TUPLES_OK)
         expect(command.cmd_tuples).to eq(2)
       end
     end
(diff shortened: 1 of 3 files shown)


andreibondarev:
@dcluna Please run `standardrb --fix` and commit. Thank you for your contribution here!

spec/langchain/vectorsearch/pgvector_spec.rb
```diff
@@ -47,7 +47,44 @@
       it "adds texts" do
         command = subject.add_texts(texts: ["Hello World", "Hello World"])
```
dcluna, line 48: The way this was written _suggests_ that ids are optional in this method - however, from the usage I could find (for instance, https://github.com/andreibondarev/langchainrb/blob/7d409a378ab300f4641a71b74d25e5cd9593122f/lib/langchain/active_record/hooks.rb#L52-L55) it seems like this is not the case.

I'd be more than happy to enforce the existence of texts/ids in tandem, but not sure if there is an (upcoming?) use case that relies on this API.

dcluna (commented):
@andreibondarev Please take a look at this comment - had to write a corner case for when ids are not required in `add_texts`, but if this is not necessary I'd prefer to have the original API using the same method for both add/update operations.

dcluna:
@andreibondarev Also, not sure if CI did not trigger due to me force-pushing the branch or if it's actually manually triggered by you - in any case, could you please make it run so this PR can be merged?

andreibondarev:
@dcluna No, at least at `add_texts()` the IDs are not required because if you don't pass IDs they should be created automatically.

PR #257, opened 2023-07-19
Stream ask() from vectorestores clients

antonzaharia:


andreibondarev:
@antonzaharia This is great, thank you! One thing to note, and I'm sure you had figured it out already, is that currently only OpenAI responses are streamed.

antonzaharia:
> @antonzaharia This is great, thank you! One thing to note, and I'm sure you had figured it out already, is that currently only OpenAI responses are streamed.

Glad you like it. Yep, OpenAI is what I used to test. :+1:

PR #259, opened 2023-07-21
Let validate_max_tokens! be inclusive to limitation

erbunao:
## Description

I noticed the validators are raising errors when the token length method returns the same length as the prescribed limitation. This PR relaxes the condition in the `validate_max_tokens!` method to allow exactly the token limitation to be used.

As a test, I can confirm sending a token equal to the published limitation on OpenAI's `text-embedding-ada-002` model went through (ie 8191 tokens). The assumption is all the published max token limitation on various LLM's API docs are actually the maximum allowed on their APIs.


andreibondarev:
@erbunao Thank you for the PR!

PR #263, opened 2023-07-29
Allow memory access

lukasedw:


lib/langchain/conversation.rb
```diff
@@ -17,18 +17,18 @@ module Langchain
...
-    attr_reader :options
+    attr_reader :options, :memory
 
     # Intialize Conversation with a LLM
     #
     # @param llm [Object] The LLM to use for the conversation
     # @param options [Hash] Options to pass to the LLM, like temperature, top_k, etc.
     # @return [Langchain::Conversation] The Langchain::Conversation instance
-    def initialize(llm:, **options, &block)
+    def initialize(llm:, memory_class: ConversationMemory, **options, &block)
```
andreibondarev, line 27: I think this is supposed to be `memory_class: BaseConversationMemory`

lib/langchain/base_conversation_memory.rb
```diff
@@ -27,10 +24,20 @@ def add_examples(examples)
...
     def append_user_message(message)
       @messages << {role: "user", content: message}
+      after_add_message(message, "user")
+    end
+
+    def after_add_message(message, role)
+      # This method can be overridden in subclasses
+    end
+
+    def end_of_conversation
```
andreibondarev, line 39: Where would this method get called?

andreibondarev:
@lukasedw I like this interface! Left you a few comments.

lib/langchain.rb
```diff
@@ -51,7 +51,7 @@ module Langchain
   autoload :Loader, "langchain/loader"
   autoload :Data, "langchain/data"
   autoload :Conversation, "langchain/conversation"
-  autoload :ConversationMemory, "langchain/conversation_memory"
+  autoload :BaseConversationMemory, "langchain/base_conversation_memory"
```
alchaplinsky, line 54: Why do we need to change the class name? I think `MyConversationMemory < Langchain::ConversationMemory` actually makes sense.

lib/langchain/conversation.rb
alchaplinsky, line 27: I'd pass a memory object instead of a class. So that the usage would be:
```ruby
llm = Langchain::LLM::OpenAI.new(...)
memory = MyCoolConversationMemory.new(llm: llm)
conversation = Langchain::Conversation(memory: memory)
```

lib/langchain/base_conversation_memory.rb
```diff
@@ -27,10 +24,20 @@ def add_examples(examples)
...
       @messages << {role: "ai", content: message}
+      after_add_message(message, "ai")
     end
 
     def append_user_message(message)
       @messages << {role: "user", content: message}
+      after_add_message(message, "user")
+    end
+
+    def after_add_message(message, role)
```
alchaplinsky, line 35: let's not create empty methods in langchain's class. It's better to do:
```ruby
def append_ai_message(content)
  append_message {role: "ai", content: content}
end

def append_message(message)
  @messages << message
end
```

and then in subclass you can override append_message:
```ruby
def append_message(message)
  super

  # do stuff here 
end
```

lib/langchain/base_conversation_memory.rb
```diff
@@ -1,10 +1,7 @@
-# frozen_string_literal: true
```
alchaplinsky, line 1: Please do not randomly remove this comment.

alchaplinsky (changes requested):

lib/langchain/conversation.rb
andreibondarev: > memory = MyCoolConversationMemory.new(llm: llm)

What would this be needed for?

I think a better approach is:
```ruby
I'd pass a memory object instead of a class. So that the usage would be:

llm = Langchain::LLM::OpenAI.new(...)
memory = DatabaseConversationMemory.new(url: "postgres://...")
conversation = Langchain::Conversation(llm: llm, memory: memory)
```

alchaplinsky: llm is used inside of ConversationMemory to calculate token limit and also to summarize history once the limit is reached. We can change that somehow, but that's not related to my comment. My comment is about passing an instance of memory to conversation instead of just class_name.

PR #266, opened 2023-08-01
Fix function calls

santib:
The bug is that the history gets broken because of the `complete_response` feature which returns a `Hash` instead of a `String` and gets stored right away in the memory. The second time a message is sent it fails:

image

`
Chat completion failed: {'id': 'chatcmpl-7ilb1RYn4tXZ92N9vdLwhG9hPZInc', 'object': 'chat.completion', 'created': 1690903747, 'model': 'gpt-3.5-turbo-0613', 'choices': [{'index': 0, 'message': {'role': 'assistant', 'content': 'Hi there! How can I assist you today?'}, 'finish_reason': 'stop'}], 'usage': {'prompt_tokens': 50, 'completion_tokens': 11, 'total_tokens': 61}} is not of type 'string' - 'messages.1.content' (Langchain::LLM::ApiError)
`

This solution is the "easiest" one in terms of changes required and backwards compatibility. But I'd consider refactoring the solution so that the LLMs objects don't return a simple string as the `response` and instead we always work with Hashes.

The `transform_messages` was adapted to match the example provided by OpenAI
image
https://openai.com/blog/function-calling-and-other-api-updates

I'd like to hear your thoughts on it before doing such change.

### DIFF
--- a/spec/langchain/llm/openai_spec.rb
+++ b/spec/langchain/llm/openai_spec.rb
@@ -151,7 +151,7 @@ RSpec.describe Langchain::LLM::OpenAI do
     let(:model) { "gpt-3.5-turbo" }
     let(:temperature) { 0.0 }
     let(:history) { [content: prompt, role: "user"] }
-    let(:parameters) { {parameters: {messages: history, model: model, temperature: temperature, max_tokens: be_between(4015, 4096)}} }
+    let(:parameters) { {parameters: {messages: history, model: model, temperature: temperature, max_tokens: be_between(4014, 4096)}} }
     let(:answer) { "As an AI language model, I don't have feelings, but I'm functioning well. How can I assist you today?" }
     let(:response) do
       {
@@ -334,7 +334,7 @@ RSpec.describe Langchain::LLM::OpenAI do
           let(:messages) do
             [
               {role: "user", content: "What is the weather like in Boston?"},
-              {role: "assistant", content: complete_response},
+              {role: "assistant", content: complete_response}
             ]
           end
           let(:history) do
--- a/spec/langchain/llm/openai_spec.rb
+++ b/spec/langchain/llm/openai_spec.rb
@@ -327,6 +327,35 @@ RSpec.describe Langchain::LLM::OpenAI do
           subject.functions = [{foo: :bar}]
           expect(subject.chat(prompt: prompt, model: model, temperature: temperature)).to be_a Hash
         end
+
+        context "with function_calls responses in the history" do
+          let(:function_call_content) { {"name" => "get_current_weather", "arguments" => {location: "Boston, MA"}.to_json} }
+          let(:complete_response) { {"choices" => [{"message" => {"content" => nil, "function_call" => function_call_content}}]} }
+          let(:messages) do
+            [
+              {role: "user", content: "What is the weather like in Boston?"},
+              {role: "assistant", content: complete_response},
+            ]
+          end
+          let(:history) do
+            [
+              {role: "user", content: "What is the weather like in Boston?"},
+              {role: "assistant", content: nil, function_call: function_call_content},
+              {role: "user", content: prompt}
+            ]
+          end
+
+          it "transforms the messages in the history correctly" do
+            subject.complete_response = true
+            subject.functions = [{foo: :bar}]
+            expect(subject.client).to receive(:chat).with(parameters).and_return(response)
+            # Check that we only send strings or null to OpenAI, otherwise it fails
+            parameters[:parameters][:messages].each do |message|
+              expect([String, NilClass]).to include(message[:content].class)
+            end
+            expect(subject.chat(prompt: prompt, messages: messages, model: model, temperature: temperature)).to be_a Hash
+          end
+        end
       end
     end
 
--- a/bin/console
+++ b/bin/console
@@ -12,7 +12,6 @@ require "langchain"
 require "dotenv/load"
 require "pry-byebug"
 
-
 weaviate = Langchain::Vectorsearch::Weaviate.new(
   url: ENV["WEAVIATE_URL"],
   api_key: ENV["WEAVIATE_API_KEY"],
--- a/lib/langchain/llm/openai.rb
+++ b/lib/langchain/llm/openai.rb
@@ -188,15 +188,24 @@ module Langchain::LLM
     end
 
     def transform_messages(messages)
-      messages.map do |message|
-        role = message[:role] || message["role"]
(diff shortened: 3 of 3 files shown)


spec/langchain/llm/openai_spec.rb
```diff
@@ -327,6 +327,35 @@
           subject.functions = [{foo: :bar}]
           expect(subject.chat(prompt: prompt, model: model, temperature: temperature)).to be_a Hash
         end
+
+        context "with function_calls responses in the history" do
```
diego-suarez, line 331: I'm not familiarized with the tool, but is the function response also added to the history after executing it on `langchainrb`? if so, maybe is worth to add a scenario that is supported that format?

I mean, openai api with function calling adds 2 new history message types, one when the model decides to call a function (the one you supported):
`{ role: "assistant", "content": null, "function_call" => "function_name" }`

But also you can add to the history an entry with `role: "function"` that has the response of the function call itself so the model knows the content of the function execution:

`{ role: "function", "content": "function execution response / response sent to the user/chat", "name": "function_name" }`

santib: AFAIK langchain.rb doesn't call functions automatically and adding that feature would mean a fair amount of work

andreibondarev:
@santib Yeah, I'm actually thinking that the LLM `complete()` response should just be a hash. Seem wrong to return the string across the board when there's more data in the hashes that's useful. I might take a quick stab at this.

andreibondarev:
@santib Would you please share the whole code to replicate this scenario?

santib:
> @santib Would you please share the whole code for me to replicate this error?

```ruby
require "dotenv/load"
require "langchain"

FUNCTIONS_SCHEMA = [{
  "name": "say",
  "description": "Say hi to someone",
  "parameters": {
    "type": "object",
    "properties": {
      "name": {
        "type": "string",
        "description": "Name of the person",
      }
    }
  }
}]

conv = Langchain::Conversation.new(llm: llm)
conv.set_functions(FUNCTIONS_SCHEMA)
conv.message('hello')
conv.message('hello')
```

spec/langchain/llm/openai_spec.rb
santib: @andreibondarev if you want we can work in https://python.langchain.com/docs/modules/chains/how_to/openai_functions in another PR

andreibondarev:
@santib It's working for me. Which gem version are you using?

santib:
> @santib It's working for me. Which gem version are you using?

🤔 strange..

langchain: 0.6.9
ruby-openai: 4.0.0

andreibondarev:
@santib Nvm -- I can replicate it now. Based on your screenshot: 
image

It seems like the proper thing to do is to also store the new `role: "function"` and `function_call:` key.

santib:
> @santib Nvm -- I can replicate it now. Based on your screenshot: image
> 
> It seems like the proper thing to do is to also store the new `role: "function"` and `function_call:` key.

but we can't do that without calling the function. I think that should be worked on in a separate PR as I said here https://github.com/andreibondarev/langchainrb/pull/266#discussion_r1281040138, since it's a bigger change. WDYT?

andreibondarev:
@santib I think we can do so incrementally, by working in tiny PRs. For example first PR could just be storing the `function_call:` key while actually calling the function and returning the `{ role: "function", name:, content: }` could be a follow up PR.

andreibondarev:
@santib I started this branch to return raw JSON responses from LLMs (OpenAI for now): https://github.com/andreibondarev/langchainrb/compare/hash-response?expand=1. What do you think?

santib:
>@santib I think we can do so incrementally, by working in tiny PRs. For example first PR could just be storing the function_call: key while actually calling the function and returning the { role: "function", name:, content: } could be a follow up PR. Thoughts?

Sounds good to me.

>@santib I started this branch to return raw JSON responses from LLMs (OpenAI for now): https://github.com/andreibondarev/langchainrb/compare/hash-response?expand=1. What do you think?

maybe you forgot to push some changes? because I don't see anything apart from refactoring   `complete_response` to `content_only`

andreibondarev:
> > @santib I think we can do so incrementally, by working in tiny PRs. For example first PR could just be storing the function_call: key while actually calling the function and returning the { role: "function", name:, content: } could be a follow up PR. Thoughts?
> 
> Sounds good to me.
> 
> >

@santib Would you have any interest in implementing this?

santib:
>@santib Would you have any interest in and time to implement this?

sure thing! I will collaborate with @diego-suarez on this 👍 

For more context: I think it's important that the Ruby Community develop AI tools (like langchain.rb - thank you) so we don't fall behind compared to other languages. I don't want to see people going with other languages just because we don't have enough tooling in this area. So yes, I'd like to help here ✌️

andreibondarev:
> > @santib I started this branch to return raw JSON responses from LLMs (OpenAI for now): https://github.com/andreibondarev/langchainrb/compare/hash-response?expand=1. What do you think?
> 
> maybe you forgot to push some changes? because I don't see anything apart from refactoring `complete_response` to `content_only`

Sorry -- this returns the JSON response by default unless you're using the `Langchain::Conversation` class.

madwire:
I've been looking at this also; I've had to patch the following methods and classes to get the Conversation to work with openai functions.

- Langchain::LLM::OpenAI#transform_messages because it strips out name and function_call params
- Langchain::Conversation#message this is because it doesn't work properly with @llm.complete_response
- Langchain::ConversationMemory#append I've added a generic append to deal with function and complete response messages
- Langchain::Conversation#function_call_response(function_name, response) I've added a way to append the function call response and run the llm again

All of my code is a bit quick and dirty to get something working. After working on the messages, I feel the project needs to create some kind of standard Message objects that can be more easily passed around and a clean way to convert it self correctly for each LLM. 

Also it might be better to split the Conversations and messages element and LLM api call from each other as I'm not sure all LLM providers will implement the function_call style the same way?

Happy to help in any way I can, just don't want to step on work others have already started.

santib:
Hey @madwire,

> After working on the messages, I feel the project needs to create some kind of standard Message objects that can be more easily passed around and a clean way to convert it self correctly for each LLM.

I agree that a Message object is needed, I guess we'll need something similar to what langchain JS has https://github.com/hwchase17/langchainjs/blob/1c1274d73b0dde9e689190fa6885b2a5046c5722/langchain/src/schema/index.ts#L329

>Happy to help in any way I can, just don't want to step on work others have already started.


No worries on my side, feel free to submit the PR ✌️

PR #276, opened 2023-08-08
Introduce messages to use instead of raw strings

santib:
This is part 1 of what was discussed here https://github.com/andreibondarev/langchainrb/pull/266

This PR also fixes the problem reported in #266 .

1. Introduce `Message` object with subclasses for each type of message, based on https://github.com/langchain-ai/langchain/blob/7fc07ba5df99b9fa8bef837b0fafa220bc5c932c/libs/langchain/langchain/schema/messages.py#L61
2. Use `Message`s instead of strings in the `chat` methods and in `Conversation`
3. Remove `complete_response` option since it became unnecessary

TODO:
- [x] Check what to do with the "summary" of the `ConverstionMemory`


lib/langchain/conversation.rb
```diff
@@ -58,10 +57,11 @@ def add_examples(examples)
...
     def message(message)
-      @memory.append_user_message(message)
-      response = llm_response(message)
-      @memory.append_ai_message(response)
-      response
+      human_message = HumanMessage.new(message)
+      @memory.append_message(human_message)
+      ai_message = llm_response(human_message)
+      @memory.append_message(ai_message)
+      ai_message
```
santib, line 64: we now return a Message object, but because it responds to `to_s`, we get the best of both worlds

lib/langchain/conversation_memory.rb
```diff
@@ -47,7 +43,7 @@ def reduce_messages(exception)
     def context
       return if @context.nil? && @summary.nil?
 
-      [@context, @summary].compact.join("\n")
+      SystemMessage.new([@context, @summary].compact.join("\n"))
     end
 
     private
```
santib, line 49: We'll need to check the `summarize_messages` method below, I'm not sure what "role" it should have.. maybe it should be a `ChatMessage` with `role: chat`?

lib/langchain/message.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+  end
+
+  class HumanMessage < Message
+    def type
+      "human"
+    end
+  end
+
+  class SystemMessage < Message
```
diego-suarez, line 47: maybe also a new message that represents a function execution, wdyt? (not used yet, so it can be on a new pr when its used):
```
FunctionMessage < Message
  def type
    "system"
  end
end
```

santib: I'd prefer to add it on the next PR so each PR is self-contained and we never have unused code

lib/langchain/conversation_memory.rb
santib: I checked, and now the difference is that we send `type: [role]` instead of `role: [role]` , but I think we might be good

lib/langchain/conversation.rb
```diff
@@ -58,10 +57,11 @@ def add_examples(examples)
     # @param message [String] The prompt to message the model with
     # @return [String] The response from the model
```
santib, line 58:
```suggestion
    # @return [AIMessage] The response from the model
```

lib/langchain/conversation.rb
```diff
@@ -39,17 +39,16 @@ def initialize(llm:, **options, &block)
...
     def set_functions(functions)
       @llm.functions = functions
-      @llm.complete_response = true
     end
 
     # Set the context of the conversation. Usually used to set the model's persona.
     # @param message [String] The context of the conversation
     def set_context(message)
-      @memory.set_context message
+      @memory.set_context SystemMessage.new(message)
```
santib, line 47: is this ok?

lib/langchain/conversation_memory.rb
```diff
@@ -47,7 +43,7 @@ def reduce_messages(exception)
     def context
       return if @context.nil? && @summary.nil?
 
-      [@context, @summary].compact.join("\n")
+      SystemMessage.new([@context, @summary].compact.join("\n"))
```
santib, line 46: is this ok?

lib/langchain/llm/google_palm.rb
```diff
@@ -146,17 +149,17 @@ def compose_chat_messages(prompt:, messages:)
...
         }
       end
     end
 
     def transform_messages(messages)
       messages.map do |message|
         {
-          author: message[:role] || message["role"],
-          content: message[:content] || message["content"]
+          author: ROLE_MAPPING[message.type] || message.type,
```
santib, line 161:
```suggestion
          author: ROLE_MAPPING.fetch(message.type, message.type),
```

lib/langchain/llm/openai.rb
```diff
@@ -189,12 +197,9 @@ def compose_chat_messages(prompt:, messages:, context:, examples:)
...
     def transform_messages(messages)
       messages.map do |message|
-        role = message[:role] || message["role"]
-        content = message[:content] || message["content"]
-
         {
-          content: content,
-          role: (role == "ai") ? "assistant" : role
+          content: message.content,
+          role: ROLE_MAPPING[message.type] || message.type
```
santib, line 202:
```suggestion
          role: ROLE_MAPPING.fetch(message.type, message.type)
```

lib/langchain/llm/openai.rb
```diff
@@ -126,16 +130,20 @@ def chat(prompt: "", messages: [], context: "", examples: [], **options)
 
       if (streaming = block_given?)
         parameters[:stream] = proc do |chunk, _bytesize|
-          yield chunk if complete_response
-          yield chunk.dig("choices", 0, "delta", "content") if !complete_response
+          delta = chunk.dig("choices", 0, "delta")
+          content = delta["content"]
+          additional_kwargs = {function_call: delta["function_call"]}.compact
+          yield AIMessage.new(content, additional_kwargs)
```
santib, line 136: does this work??

santib: confirmed it works if `Langchain::AIMessage` is used 🤣 fixed ✅

lib/langchain/message.rb
```diff
@@ -0,0 +1,53 @@
...
+        content: content
+      }
+
+      hash[:additional_kwargs] = additional_kwargs unless additional_kwargs.nil? || additional_kwargs.empty?
+
+      hash.to_json
+    end
+  end
+
+  class AIMessage < Message
```
andreibondarev, line 36: I think ideally we'd split the classes into separate files.

lib/langchain/llm/google_palm.rb
```diff
@@ -104,7 +109,7 @@ def chat(prompt: "", messages: [], context: "", examples: [], **options)
       response = client.generate_chat_message(**default_params)
       raise "GooglePalm API returned an error: #{response}" if response.dig("error")
 
-      response.dig("candidates", 0, "content")
+      Langchain::AIMessage.new(response.dig("candidates", 0, "content"))
```
alchaplinsky, line 112: Why do we wrap low-level LLM API response into a higher-level `Message` class? I believe `AIMessage`/`Message` should be a part of `Conversation` higher-level abstraction. LLM classes should not return a `Message` instance.
cc @santib @andreibondarev

santib: that's a good point.. I did it here because I don't want to have a place like `Conversation` know how to access each LLM response keys, but maybe we can make use of another class so that neither `Conversation` nor `LLM::[x]` build the message? 🤔

At the same time, if we do that we'll go back to the original problem: the LLM responding just a string (which doesn't work fine with function calls)

santib: For reference, Python builds an `LLMResult` object https://github.com/langchain-ai/langchain/blob/3adb1e12cae290a4fe6500bed04e07219d652e74/libs/langchain/langchain/llms/openai.py#L876-L881

maybe that's what we need here

andreibondarev: Yeah, but they also separate the chat functionality into [langchain/chat_models/openai.py](https://github.com/langchain-ai/langchain/blob/3adb1e12cae290a4fe6500bed04e07219d652e74/libs/langchain/langchain/chat_models/openai.py).

In general I'd like to avoid introducing abstraction quickly because they're hard to get rid of afterwards.

So we know that we should change the LLM::* responses to return full hash responses, this has been mentioned by different people many times.

I agree with @alchaplinsky here I think the `LLM::OpenAI#chat()` should just return a hash and then we'd convert it to an AIMessage here: https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/conversation.rb#L62, just like it's done with the HumanMessage.

santib: but if we do that, we'll need to have `Conversation` know how to get the "content" and potentially other stuff (e.g. "function_call") from the different LLMs, right?

andreibondarev: Shouldn't the Conversation class be an encapsulation into complete functionality to have a conversation with an LLM? That's why I think the hash responses should be converted to Messages there. I think having Parsers or some sort of a hash lookup to remap is fine.

santib: I can take a stab at this refactor. I'd like to validate if this decoupling of `Message`s with `LLM`s apply only on the return value, or if it should also apply to the `prompt` (go back to `String` instead of a `Message`), and what happen with `messages` and `examples` which in the case of messages they are tailored for a conversation.

I'm fine changing the prompt + return value, but for messages and examples I think it'd be better to work with objects rather than hashes than need to conform to a format like `{ role: .., content: .. }`

PR #277, opened 2023-08-08
Run tests in random order to prevent bigger issues in the future

santib:
I think it's a [best practice to run tests in random order](https://github.com/rspec/rspec-core/pull/2929) to prevent dependencies between them.

BTW when running tests in this branch, it took only 2 runs to surface an issue
```
1) Langchain::LLM::OpenAI#initialize when llm_options are passed passes correct options to the client
     Failure/Error: expect(OpenAI.configuration.uri_base).to eq("http://localhost:1234")

       expected: "http://localhost:1234"
            got: "https://api.openai.com/"

       (compared using ==)
     # ./spec/langchain/llm/openai_spec.rb:22:in `block (4 levels) in <top (required)>'
```


andreibondarev:
@santib Nice find, and fix!

PR #278, opened 2023-08-08
Use Zeitwerk

santib:
I think there is no need to manually require/autoload things when we have a gem such as Zeitwerk in our language ❤️


alchaplinsky:
Cool. I like it!

lukasedw:
Omg, please merge it!

PR #281, opened 2023-08-09
Add destroy_default_schema to PGVector

santib:
I added this missing method because I was needing it to re-generate the table after that 0.6.8 -> 0.6.9 upgrade


lib/langchain/vectorsearch/pgvector.rb
```diff
@@ -90,7 +90,6 @@ def update_texts(texts:, ids:)
     end
 
     # Create default schema
-    # @return [PG::Result] The response from the database
```
santib, line 93: For me this returns `nil` not `PG::Result` 🤔

PR #282, opened 2023-08-09
Update usage example of PGVector

santib:


lib/langchain/vectorsearch/pgvector.rb
```diff
@@ -93,13 +93,12 @@ def update_texts(texts:, ids:)
     # @return [PG::Result] The response from the database
     def create_default_schema
       db.run "CREATE EXTENSION IF NOT EXISTS vector"
-      namespace = namespace_column
```
santib, line 96: no need to do this and shadow the `namespace` attr_reader 😬

andreibondarev:
@santib The specs are failing.

lib/langchain/vectorsearch/pgvector.rb
santib: well it seems that the local variable is needed indeed, but I changed it a bit so to prevent confusion between namespace and namespace_column

andreibondarev: Isn't `@namespace_column` just going to be `nil` since it's not set anywhere?

santib: it is set in the `initialize` method

PR #284, opened 2023-08-09
Decouple LLM from Messages

santib:
What should an LLM return?
- [x] The full JSON response (e.g. `{ "choices" => [{ "content" => nil, "function_call": { ... } }, {...}] }`)
- [ ] A minimally processed JSON like `{ "content" => nil, "function_call": { ... } }`
- [ ] An Object containing `content` and `extra` attributes
- [ ] A string

What should a Conversation return?
- [ ] The full JSON response (e.g. `{ "choices" => [{ "content" => nil, "function_call": { ... } }, {...}] }`)
- [ ] A minimally processed JSON like `{ "content" => nil, "function_call": { ... } }`
- [x] An Object containing `content` and `extra` attributes
- [ ] A string

Some history:
- They both used to return a String, but that didn't support OpenAI's function_call feature
- They are currently returning an Object with content and other attributes, but LLMs should be lower level


andreibondarev:
@santib I would think:

> What should an LLM return?

The full JSON response

> What should a Conversation return?

An Object containing content and extra attributes

But I'm open to other ideas!

lib/langchain/llm/google_palm.rb
```diff
@@ -75,20 +75,20 @@ def complete(prompt:, **params)
     #
     # Generate a chat completion for a given prompt
     #
-    # @param prompt [HumanMessage] The prompt to generate a chat completion for
+    # @param prompt [String] The prompt to generate a chat completion for
     # @param messages [Array<AIMessage|HumanMessage>] The messages that have been sent in the conversation
```
santib, line 79: is it ok to use Message objects instead of structured Hashes for `messages` and `examples` ?

lib/langchain/message.rb
```diff
@@ -34,6 +34,20 @@ def to_json(options = {})
   end
 
   class AIMessage < Message
+    def self.from_llm_response(llm_klass, llm_response)
```
santib, line 37: is it fine that Message know how to deal with each LLM?

santib:
@alchaplinsky thoughts on these changes?

alchaplinsky:
@santib I'm going to take a closer look. However, I think having llm-specific parsing in `Message` class is not great. Because anyone who adds a new LLM integration will definitely miss that they need to extend this method on a `Message` class.

lib/langchain/message.rb
andreibondarev: I guess another approach is to have each LLM specify a `completion_response_path` constant, something like the following:
```ruby
class Langchain::LLM::OpenAI
   COMPLETION_RESPONSE_PATH = ["choices", 0, "message"]

class Langchain::LLM::GooglePalm
   COMPLETION_RESPONSE_PATH = ["candidates", 0, "content"]
```
and then here you could just do:
```ruby
new llm_response.dig(**llm_klass::COMPLETION_RESPONSE_PATH)
```

lib/langchain/llm/google_palm.rb
andreibondarev: So maybe the Message objects are only known to the Conversation class but not the LLMs. Wdyt?

I think it's weird if we use the Message objects in _some_ places in the Conversation classes and in _some_ places in the LLM::* classes. It should either be used everywhere, always, or a clean separation where when a class knows about Messages it uses them everywhere in that class but doesn't randomly expose it to other classes.

andreibondarev:
@santib @alchaplinsky See my comment here: https://github.com/andreibondarev/langchainrb/pull/284/files#r1291771657

lib/langchain/conversation.rb
```diff
@@ -84,11 +85,44 @@ def examples
...
       retry
     end
+
+    def completion_response_path
+      key = @block ? "STREAMING_RESPONSE_PATH" : "COMPLETION_RESPONSE_PATH"
+
+      safe_const_get(llm_klass, key)
+    end
+
+    def transform_messages(messages)
```
santib, line 101: If we really want to have the LLMs know nothing about other stuff, the `transform_messages` method needs to be moved to the Conversation, so that LLMs already receive messages in the corresponding format.

lib/langchain/llm/openai.rb
```diff
@@ -18,10 +18,12 @@ class OpenAI < Base
...
-    ROLE_MAPPING = {
-      "ai" => "assistant",
-      "human" => "user"
-    }
+    COMPLETION_RESPONSE_PATH = ["choices", 0, "message"]
+    STREAMING_RESPONSE_PATH = ["choices", 0, "delta"]
+    ROLE_KEY = :role
+    USER_ROLE_KEY = "user"
+    AI_ROLE_KEY = "assistant"
+    SYSTEM_ROLE_KEY = "system"
```
santib, line 26: this is starting to get strange. The problem is the conflict between:
- We don't want the LLMs to know about anything else
- We can't put code in other places like `Message` or `Conversation` because when adding more LLMs we'll forget about those

andreibondarev: > We can't put code in other places like Message or Conversation because when adding more LLMs we'll forget about those

This one is not a valid issue. If this ever happens -- it just means that our test coverage is bad.

santib: agree.. then maybe we just need to accept having the Conversation know about each LLM specifics, instead of having each LLM having to provide things in this way (indirectly knowing that someone needs these things to be defined).

And with that we go back to the solution where Message (or could be Conversation or another class in the same "domain") knows how to parse an OpenAI result, Google Palm 2 result, etc

andreibondarev:
Currently:
```ruby
irb(main):007:0> chat.messages
=>
[#<Langchain::HumanMessage:0x00000001116c40a0 @additional_kwargs=nil, @content="Hey!">,
 #<Langchain::AIMessage:0x00000001116f47f0 @additional_kwargs={"role"=>"assistant"}, @content="Hello! How can I assist you today?">,
 #<Langchain::HumanMessage:0x0000000110c12af8 @additional_kwargs=nil, @content="Please help me debug my computer">,
 #<Langchain::AIMessage:0x0000000110bc1b30
  @additional_kwargs={"role"=>"assistant"},
  @content="Of course! I'll do my best to help you. Could you please provide more details about the issue you're experiencing? What seems to be the problem with your computer?">]
```

@santib I'm wondering if the `role:` should be an instance var on it's own along with the @content. So that each class always has a `@role` and `@content`.

santib:
> Currently:
> 
> ```ruby
> irb(main):007:0> chat.messages
> =>
> [#<Langchain::HumanMessage:0x00000001116c40a0 @additional_kwargs=nil, @content="Hey!">,
>  #<Langchain::AIMessage:0x00000001116f47f0 @additional_kwargs={"role"=>"assistant"}, @content="Hello! How can I assist you today?">,
>  #<Langchain::HumanMessage:0x0000000110c12af8 @additional_kwargs=nil, @content="Please help me debug my computer">,
>  #<Langchain::AIMessage:0x0000000110bc1b30
>   @additional_kwargs={"role"=>"assistant"},
>   @content="Of course! I'll do my best to help you. Could you please provide more details about the issue you're experiencing? What seems to be the problem with your computer?">]
> ```
> 
> @santib I'm wondering if the `role:` should be an instance var on it's own along with the @content. So that each class always has a `@role` and `@content`.

1. I need to check why `role` is present in the `additional_kwargs` because that was not the idea 🤔 
2. We could make `role` be an attr same as `content`. I made different classes in part to follow what was done in Python/JS Langchain implementations. The other reason is that having different classes, make the interface more clear (`@param message [Message] a message with role human` vs `@param message [HumanMessage] a message`). That said, I'm good with whatever option you prefer

andreibondarev:
> > Currently:
> > ```ruby
> > irb(main):007:0> chat.messages
> > =>
> > [#<Langchain::HumanMessage:0x00000001116c40a0 @additional_kwargs=nil, @content="Hey!">,
> >  #<Langchain::AIMessage:0x00000001116f47f0 @additional_kwargs={"role"=>"assistant"}, @content="Hello! How can I assist you today?">,
> >  #<Langchain::HumanMessage:0x0000000110c12af8 @additional_kwargs=nil, @content="Please help me debug my computer">,
> >  #<Langchain::AIMessage:0x0000000110bc1b30
> >   @additional_kwargs={"role"=>"assistant"},
> >   @content="Of course! I'll do my best to help you. Could you please provide more details about the issue you're experiencing? What seems to be the problem with your computer?">]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @santib I'm wondering if the `role:` should be an instance var on it's own along with the @content. So that each class always has a `@role` and `@content`.
> 
> 1. I need to check why `role` is present in the `additional_kwargs` because that was not the idea 🤔
> 2. We could make `role` be an attr same as `content`. I made different classes in part to follow what was done in Python/JS Langchain implementations. The other reason is that having different classes, make the interface more clear (`@param message [Message] a message with role human` vs `@param message [HumanMessage] a message`). That said, I'm good with whatever option you prefer

1. Makes sense!
2. Yeah, the different classes: HumanMessage, AIMessage make sense to me.

If you're able to wrap up this PR -- that would be great! I think this is the first PR that needs to go in before all others right now.

lib/langchain/conversation.rb
```diff
@@ -84,11 +85,44 @@ def examples
...
+      end
+    end
+
+    def role_mapping
+      @role_mapping ||= {
+        "ai" => safe_const_get(llm_klass, "AI_ROLE_KEY"),
+        "human" => safe_const_get(llm_klass, "USER_ROLE_KEY"),
+        "system" => safe_const_get(llm_klass, "SYSTEM_ROLE_KEY")
+      }.compact
+    end
```
santib, line 118: @andreibondarev I just rebased the PR and fixed the thing about the role getting into the additional_kwargs. This PR is ready to go unless we want to do something with the constants defined in LLMs (`ROLE_KEY`,  `AI_ROLE_KEY`, etc). Do we want to leave it like this or have the Conversation know about the LLMs` response formats?

lib/langchain/llm/google_palm.rb
```diff
@@ -19,9 +19,9 @@ class GooglePalm < Base
       embeddings_model_name: "embedding-gecko-001"
     }.freeze
     LENGTH_VALIDATOR = Langchain::Utils::TokenLength::GooglePalmValidator
-    ROLE_MAPPING = {
-      "human" => "user"
-    }
+    COMPLETION_RESPONSE_PATH = ["candidates", 0]
+    ROLE_KEY = :author
```
alchaplinsky, line 23: I don't think we need to have these role mappings at all. This is only needed in the context of conversation and `Conversation` class is already aware if the message was sent by the user it is a response from LLM.

andreibondarev:
@santib The specs are broken currently.

andreibondarev:
@santib I'm deprecating the whole `Langchain::Conversation` class and introducing a new `Langchain::Assistant` class here: https://github.com/andreibondarev/langchainrb/pull/422 and closing this PR in the meantime. Thank you for all of the work and constructive back and forth conversations that did take place here!

PR #285, opened 2023-08-09
Improve functions

lukasedw:
The motivation is to fully [OpenAI functions](https://openai.com/blog/function-calling-and-other-api-updates)

### DIFF
--- a/lib/langchain/conversation.rb
+++ b/lib/langchain/conversation.rb
@@ -64,6 +64,18 @@ module Langchain
       ai_message
     end
 
+    # Message the model with the function name, content and return the response.
+    # @param message [String] The content or arguments for the function call
+    # @param name [String] The name of the function to be called
+    # @return [String] The response from the function call
+    def function(message, name)
+      function_message = FunctionMessage.new(message, {name: name})
+      @memory.append_message(function_message)
+      ai_message = llm_response
+      @memory.append_message(ai_message)
+      ai_message
+    end
+
     # Messages from conversation memory
     # @return [Array<AIMessage|HumanMessage>] The messages from the conversation memory
     def messages
--- a/lib/langchain/conversation.rb
+++ b/lib/langchain/conversation.rb
@@ -59,7 +59,7 @@ module Langchain
     def message(message)
       human_message = HumanMessage.new(message)
       @memory.append_message(human_message)
-      ai_message = llm_response(human_message)
+      ai_message = llm_response
       @memory.append_message(ai_message)
       ai_message
     end
@@ -84,7 +84,7 @@ module Langchain
 
     private
 
-    def llm_response(prompt)
+    def llm_response
       @llm.chat(messages: @memory.messages, context: @memory.context, examples: @memory.examples, **@options, &@block)
     rescue Langchain::Utils::TokenLength::TokenLimitExceeded => exception
       @memory.reduce_messages(exception)
--- a/lib/langchain/conversation.rb
+++ b/lib/langchain/conversation.rb
@@ -17,7 +17,7 @@ module Langchain
   #     end
   #
   class Conversation
-    attr_reader :options
+    attr_reader :options, :memory
 
     # Intialize Conversation with a LLM
     #
--- a/lib/langchain/llm/openai.rb
+++ b/lib/langchain/llm/openai.rb
@@ -194,7 +194,7 @@ module Langchain::LLM
         {
           role: ROLE_MAPPING.fetch(message.type, message.type),
           content: message.content
-        }
+        }.merge(message.additional_kwargs || {})
       end
     end
 
--- a/lib/langchain/message.rb
+++ b/lib/langchain/message.rb
@@ -50,4 +50,10 @@ module Langchain
       "system"
     end
   end
+
+  class FunctionMessage < Message
+    def type
+      "function"
+    end
+  end
 end


lib/langchain/conversation.rb
```diff
@@ -17,7 +17,7 @@ module Langchain
   #     end
   #
   class Conversation
-    attr_reader :options
+    attr_reader :options, :memory
```
lukasedw, line 20: WIth this, it will open possibilities to manipulate the memory as the dev want

lib/langchain/llm/openai.rb
```diff
@@ -192,7 +192,7 @@ def compose_chat_messages(prompt:, messages:, context:, examples:)
     def transform_messages(messages)
       messages.map do |message|
         {
-          role: ROLE_MAPPING.fetch(message.type, message.type),
+          role: ROLE_MAPPING.fetch(message.type) { raise "Role not found for #{message.type}" },
```
lukasedw, line 195: Raising here will prevent tokens from being spent unnecessarily.

lib/langchain/conversation.rb
```diff
@@ -59,7 +59,19 @@ def add_examples(examples)
...
+      ai_message = llm_response
+      @memory.append_message(ai_message)
+      ai_message
+    end
+
+    # Message the model with the function name, content and return the response.
+    # @param message [String] The content or arguments for the function call
+    # @param name [String] The name of the function to be called
+    # @return [String] The response from the function call
+    def function(message, name)
```
lukasedw, line 71: I think we should have this method coming from LLMs, but it is here for feedback

andreibondarev:
@santib Would you mind reviewing this PR please?

lib/langchain/conversation.rb
santib: I'm not sure about this change. Even if we wanted to do this, I think it could be discussed separately and doesn't need to be part of this PR. Or am I missing something?

lib/langchain/conversation.rb
```diff
@@ -59,7 +59,19 @@ def add_examples(examples)
...
       @memory.append_message(human_message)
-      ai_message = llm_response(human_message)
+      ai_message = llm_response
+      @memory.append_message(ai_message)
+      ai_message
+    end
+
+    # Message the model with the function name, content and return the response.
+    # @param message [String] The content or arguments for the function call
+    # @param name [String] The name of the function to be called
```
santib, line 69: If I understand these comments correctly (e.g. "the function to be called") you expect that the LLM somehow invoke the function, am I right?


In [step 3 of this example](https://openai.com/blog/function-calling-and-other-api-updates) we can see a function message and it looks like
`{"role": "function", "name": "get_current_weather", "content": "{\"temperature\": "22", \"unit\": \"celsius\", \"description\": \"Sunny\"}"}`

which I understand as a function that was already invoked and the content carries its result.

lib/langchain/conversation.rb
```diff
@@ -59,7 +59,19 @@ def add_examples(examples)
...
+    # @return [String] The response from the function call
+    def function(message, name)
+      function_message = FunctionMessage.new(message, {name: name})
+      @memory.append_message(function_message)
+      ai_message = llm_response
```
santib, line 74: so, is the idea that every time we invoke a function, we then invoke the LLM to convert the result to a more friendly result?

I was thinking that it might not be always necessary and instead we need a more dumb API that just lets us add the result of having called the function that was suggested by the LLM, something like https://github.com/andreibondarev/langchainrb/pull/286/files#diff-9672ba0f87466bc23c7508052b5e8f37280c2ec8efd46faecacf4c9d72f6c24dR56-R61 WDYT?

lib/langchain/conversation.rb
```diff
@@ -84,7 +96,7 @@ def examples
 
     private
 
-    def llm_response(prompt)
+    def llm_response
```
santib, line 99: 💪

lib/langchain/conversation.rb
lukasedw: It is convenient when someone (like me) wants to manipulate memory for various purposes.

lib/langchain/conversation.rb
lukasedw: Not exactly the result, but what you want the LLM to understand. For example, if you write on the content "API error", the LLM will process and generate a message to the user. So, it depends on the case.

lib/langchain/conversation.rb
lukasedw: Why would it not be necessary? This is a real necessity of an MVP product. The flow goes like this:

user: I want to make an appointment

Here it gets interesting. Sometimes OpenAI functions call the function with empty or nonsense arguments. Then with this method is possible to "double-check":

``` ruby
chat.function("The appointment time is empty", "appointment")
```
#### Get the info like time, day, and so on
ai:  call function appointment with arguments
#### Everything is valid:
``` ruby
chat.function("Confirm the appointment", "appointment")
```
#### The function process and give the result:

``` ruby
chat.function("Appointment confirmed", "appointment")
```

It could be solved with different approaches, but it is just an example.

So, if we just append using that method and don't call the LLM to process the result, doesn't make sense.

lukasedw:
Any thoughts here @andreibondarev? My only concern is where the function method will live on.

PR #286, opened 2023-08-10
Function calling

santib:
Not sure if an Agent is fine or if it should be something else, but the idea is to implement the example here https://openai.com/blog/function-calling-and-other-api-updates

This work with the existing tools but you can create your own tools and pass them too. e.g.

```ruby
class SayHi < Langchain::Tool::Base
  NAME = "say_hi"

  description <<~DESC
    Say hi to someone
  DESC

  SCHEMA_PARAMETERS = {
    name: {
      type: "string",
      description: "Name of the person to say hi"
    }
  }

  def execute(name:)
    puts "Hi, #{name}!"
  end
end
``` 

Working example:
```
irb(main):003:0> agent = Langchain::Agent::FunctionCallAgent.new(llm: llm, tools: [L
angchain::Tool::Calculator.new])
=>
#<Langchain::Agent::FunctionCallAgent:0x0000000109602cc8
...
irb(main):005:0> agent.run(question: "2 x 5")
=> 10.0
irb(main):006:0> agent.run(question: "what was the previous result?")
=> "The previous result was 10."
```

If this is something we want to add then I can add tests

### DIFF
--- a/lib/langchain/agent/function_call_agent/function_call_agent.rb
+++ b/lib/langchain/agent/function_call_agent/function_call_agent.rb
@@ -29,7 +29,9 @@ module Langchain::Agent
 
       if (function = ai_response.additional_kwargs[:function_call])
         tool = tools.find { |tool| tool.tool_name == function["name"] }
-        tool.execute(**JSON.parse(function["arguments"]).transform_keys(&:to_sym))
+        tool.execute(**JSON.parse(function["arguments"]).transform_keys(&:to_sym)).tap do |result|
+          conversation.add_function_call_result(function["name"], result)
+        end
       else
         ai_response.content
       end
--- a/lib/langchain/agent/function_call_agent/function_call_agent.rb
+++ b/lib/langchain/agent/function_call_agent/function_call_agent.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module Langchain::Agent
+  class FunctionCallAgent < Base
+    attr_reader :conversation, :tools
+
+    #
+    # Initializes the Agent
+    #
+    # @param llm [Object] The LLM client to use
+    # @param tools [Array<Object>] The tools that can be invoked
+    #
+    def initialize(llm:, tools:)
+      Langchain::Tool::Base.validate_tools!(tools: tools)
+
+      @conversation = Langchain::Conversation.new(llm: llm)
+      @tools = tools
+    end
+
+    #
+    # Ask a question and responds or call the appropriate function
+    #
+    # @param question [String] Question to ask the LLM
+    # @return [String] Answer to the question
+    #
+    def run(question:)
+      conversation.set_functions(tools.map(&:tool_schema))
+      ai_response = conversation.message(question)
+
+      if (function = ai_response.additional_kwargs[:function_call])
+        tool = tools.find { |tool| tool.tool_name == function["name"] }
+        tool.execute(**JSON.parse(function["arguments"]).transform_keys(&:to_sym))
+      else
+        ai_response.content
+      end
+    end
+  end
+end
--- a/lib/langchain/agent/function_call_agent/function_call_agent.rb
+++ b/lib/langchain/agent/function_call_agent/function_call_agent.rb
@@ -24,11 +24,11 @@ module Langchain::Agent
     # @return [String] Answer to the question
     #
     def run(question:)
-      conversation.set_functions(tools.map(&:tool_schema))
+      conversation.set_functions(tools.map(&:schema))
       ai_response = conversation.message(question)
 
       if (function = ai_response.additional_kwargs[:function_call])
-        tool = tools.find { |tool| tool.tool_name == function["name"] }
+        tool = tools.find { |tool| tool.name == function["name"] }
         tool.execute(**JSON.parse(function["arguments"]).transform_keys(&:to_sym)).tap do |result|
           conversation.add_function_call_result(function["name"], result)
         end
--- a/lib/langchain/conversation.rb
+++ b/lib/langchain/conversation.rb
@@ -53,6 +53,13 @@ module Langchain
       @memory.add_examples examples
     end
 
+    # Add function call result to the conversation.
+    # @param name [String] The name of the function
+    # @param result [Any] The result of a function call
+    def add_function_call_result(name, result)
+      @memory.append_message(FunctionMessage.new(result.to_json, { "name" => name }))
+    end
+
     # Message the model with a prompt and return the response.
     # @param message [String] The prompt to message the model with
     # @return [AIMessage] The response from the model
--- a/lib/langchain/conversation.rb
+++ b/lib/langchain/conversation.rb
@@ -57,7 +57,7 @@ module Langchain
     # @param name [String] The name of the function
     # @param result [Any] The result of a function call
     def add_function_call_result(name, result)
-      @memory.append_message(FunctionMessage.new(result.to_json, { "name" => name }))
+      @memory.append_message(FunctionMessage.new(result.to_json, {"name" => name}))
     end
 
     # Message the model with a prompt and return the response.
--- a/lib/langchain/tool/base.rb
+++ b/lib/langchain/tool/base.rb
@@ -77,10 +77,10 @@ module Langchain::Tool
     #
     # @return [Hash] tool description
     #
-    def tool_schema
+    def schema
       {
-        name: tool_name,
-        description: tool_description,
+        name: name,
+        description: description,
         parameters: {
           type: "object",
           properties: self.class.const_get(:SCHEMA_PARAMETERS)
--- a/lib/langchain/tool/base.rb
+++ b/lib/langchain/tool/base.rb
@@ -73,6 +73,22 @@ module Langchain::Tool
     end
 
     #
+    # Returns the SCHEMA constant of the tool
+    #
+    # @return [Hash] tool description
+    #
+    def tool_schema
+      {
+        name: tool_name,
+        description: tool_description,
+        parameters: {
+          type: "object",
+          properties: self.class.const_get(:SCHEMA_PARAMETERS)
+        }
+      }
+    end
+
+    #
     # Sets the DESCRIPTION constant of the tool
     #
     # @param value [String] tool description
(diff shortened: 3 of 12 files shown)


lib/langchain/agent/function_call_agent/function_call_agent.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module Langchain::Agent
+  class FunctionCallAgent < Base
```
santib, line 4: not sure about this name, it's a conversation + function caller

lib/langchain/tool/base.rb
```diff
@@ -72,6 +72,22 @@ def tool_description
       self.class.const_get(:DESCRIPTION)
     end
 
+    #
+    # Returns the SCHEMA constant of the tool
+    #
+    # @return [Hash] tool description
+    #
+    def tool_schema
```
diego-suarez, line 80:
```suggestion
    def json_schema
```
wdyt?

lib/langchain/tool/base.rb
```diff
@@ -72,6 +72,22 @@ def tool_description
...
+      {
+        name: tool_name,
+        description: tool_description,
```
diego-suarez, line 83: not related to this pr, but maybe rename tool_name and tool_description to just name and description? looks redundant to me the `tool_` prefix. thoughts @andreibondarev ?

lib/langchain/tool/database.rb
```diff
@@ -14,6 +14,13 @@ class Database < Base
       The input to this tool should be valid SQL.
     DESC
 
+    SCHEMA_PARAMETERS = {
+      input: {
+        type: "string",
+        description: "SQL expression"
+      }
```
diego-suarez, line 21: I think this change will be a bit more informative to the LLM:
```suggestion
      sql_query: {
        type: "string",
        description: "SQL query"
      }
```
but then need to rename the argument on the `execute` method too. Will that rename break original tools implementation? maybe can be an enhancement for other pr

lib/langchain/tool/ruby_code_interpreter.rb
```diff
@@ -8,10 +8,18 @@ class RubyCodeInterpreter < Base
...
+
     description <<~DESC
       A Ruby code interpreter. Use this to execute ruby expressions. Input should be a valid ruby expression. If you want to see the output of the tool, make sure to return a value.
     DESC
 
+    SCHEMA_PARAMETERS = {
+      input: {
+        type: "string",
+        description: "Ruby code expression"
+      }
```
diego-suarez, line 20: same here 
```suggestion
      code: {
        type: "string",
        description: "executable Ruby code"
      }
```

lib/langchain/tool/weather.rb
```diff
@@ -30,6 +30,13 @@ class Weather < Base
         Action Input: Kiev, Ukraine; metric
     DESC
 
+    SCHEMA_PARAMETERS = {
+      input: {
+        type: "string",
+        description: "Comma separated city and unit (optional: imperial, metric, or standard)"
+      }
```
diego-suarez, line 37: now that supports multiple arguments, this can be better to split in 2, city and unit. with city required, and unit optional

lib/langchain/tool/base.rb
```diff
@@ -72,6 +72,22 @@ def tool_description
...
+        parameters: {
+          type: "object",
+          properties: self.class.const_get(:SCHEMA_PARAMETERS)
+        }
+      }
```
diego-suarez, line 88: wdyt about implementing requiered parameters? they are defined inside parameters object, and is an array of strings. possible implementation:

```
{
        name: tool_name,
        description: tool_description,
        parameters: {
          type: "object",
          properties: self.class.const_get(:SCHEMA_PARAMETERS),
          required: self.class.const_get(:REQUIRED_PARAMETERS)
        },
      }
```
and then each tool defines its own `REQUIRED_PARAMETERS` const

diego-suarez (approved):
awesome 🤩

lib/langchain/agent/function_call_agent/function_call_agent.rb
```diff
@@ -0,0 +1,40 @@
...
+    # @return [String] Answer to the question
+    #
+    def run(question:)
+      conversation.set_functions(tools.map(&:tool_schema))
+      ai_response = conversation.message(question)
+
+      if (function = ai_response.additional_kwargs[:function_call])
+        tool = tools.find { |tool| tool.tool_name == function["name"] }
+        tool.execute(**JSON.parse(function["arguments"]).transform_keys(&:to_sym)).tap do |result|
+          conversation.add_function_call_result(function["name"], result)
```
santib, line 33: should we call the llm one more time to get a more friendly answer to the user?

diego-suarez: what do you think to leave that to the tool? If the tool wants to use an llm to get the answer then it can. I dont know if I like the idea of all responses to be through the llm (making an extra request, adding cost / delay)

lib/langchain/agent/function_call_agent/function_call_agent.rb
```diff
@@ -0,0 +1,40 @@
...
+  class FunctionCallAgent < Base
+    attr_reader :conversation, :tools
+
+    #
+    # Initializes the Agent
+    #
+    # @param llm [Object] The LLM client to use
+    # @param tools [Array<Object>] The tools that can be invoked
+    #
+    def initialize(llm:, tools:)
```
diego-suarez, line 13: should we validate here if the llm supports openai? if llm is not that particular one it probably will fail with an unexpected error right? maybe is better to be proactive here so its clear to the developer the error, and don't need to debug whats going on/why is failing

diego-suarez: another option is to not receive the llm but create it on the agent 🤔
I don't know about other models supporting function calling, but probably will not be an exact implementation, so if that's true, for now maybe is worth to make the implementation specific to openai api?

lib/langchain/tool/base.rb
santib: I don't like it that much, I think it's better to keep it consistent with the rest of the names. Maybe if your last PR gets approved this will be just `schema`

lib/langchain/tool/database.rb
santib: maybe a change like this could break non-function calling flows. I mean, using tools without function calling

lib/langchain/agent/function_call_agent/function_call_agent.rb
santib: I get your point, but I think it's expected that agents do this 🤔 for example it is done in the `SQLQueryAgent`

lib/langchain/agent/function_call_agent/function_call_agent.rb
santib: FTR, this is the current behavior:
```ruby
irb(main):003:0> a = Langchain::Agent::FunctionCallAgent.new(llm: Langchain::LLM::AI21.new(api_key: '123'), tools: [Langchain::Tool::Calculator.new])
=>
#<Langchain::Agent::FunctionCallAgent:0x000000010362f008
...
irb(main):004:0> a.run(question: 'a')
/Users/user/dev/langchainrb/lib/langchain/conversation.rb:41:in `set_functions': undefined method `functions=' for #<Langchain::LLM::AI21:0x00000001040d2d20 @client=#<AI21::Client:0x00000001042d0ac8>, @defaults={:temperature=>0.0, :model=>"j2-ultra"}> (NoMethodError)```

File: lib/langchain/conversation.rb [hunk 10]
```diff
@@ -53,6 +53,13 @@ def add_examples(examples)
       @memory.add_examples examples
     end
 
+    # Add function call result to the conversation.
+    # @param name [String] The name of the function
+    # @param result [Any] The result of a function call
+    def add_function_call_result(name, result)
+      @memory.append_message(FunctionMessage.new(result.to_json, {"name" => name}))
```
lukasedw, line 60: The result don't need to always be a json.

andreibondarev:
@santib What are your thoughts on baking this into the `Langchain::Conversation` class?

For example we could add the `tools:` parameter:
```ruby
Langchain::Conversation.new llm:, tools: [Langchain::Calculator.new]
```

The Conversation class could automatically reflect on all of the `execute()` methods within the Tools and generate the required functions JSON:
```ruby
# something like... but could get more or less creative:
functions = generate_functions_json(
  parameters: Langchain::Tool::Calculator.method(:execute).parameters,
  description: Langchain::Tool::Calculator.method(:execute).description
)
```

santib:
> @santib What are your thoughts on baking this into the `Langchain::Conversation` class?
> 
> For example we could add the `tools:` parameter:
> 
> ```ruby
> Langchain::Conversation.new llm:, tools: [Langchain::Calculator.new, ...]
> ```
> 
> The Conversation class could automatically reflect on all of the `execute()` methods within the Tools and generate the required functions JSON:
> 
> ```ruby
> # something like... but could get more or less creative:
> functions = generate_functions_json(
>   parameters: Langchain::Tool::Calculator.method(:execute).parameters,
>   description: Langchain::Tool::Calculator.method(:execute).description
> )
> ```

@andreibondarev I'm not sure it makes sense from a Langchain perspective, but that might not matter if we are going to diverge from the original idea.

Apart from that, if we add tools to conversations, will it only work with LLM's "function calling" feature (only available in OpenAI for now) or are we also going to do something more custom like it's done in https://github.com/andreibondarev/langchainrb/tree/main/lib/langchain/agent/react_agent/react_agent_prompt.yaml when it doesn't support "function calling" natively?

andreibondarev:
@santib Yeah, we'd need to do something custom for OpenAI.  Btw -- do you have an actual real use-case you'd like to solve here?

santib:
> @santib Yeah, we'd need to do something custom for OpenAI. Btw -- do you have an actual real use-case you'd like to solve here?

Not for now, it's mostly trying to stay up to date with new tooling available so Ruby devs can compete with devs from other languages. Personally speaking, I don't care if it's langchain (I don't actually like their mental model) or any other AI framework, but we need to provide Ruby devs with powerful tools so they can smoothly augment their apps with AI.

andreibondarev:
`function_call:` and `functions:` are deprecated (https://platform.openai.com/docs/api-reference/chat/create#chat-create-function_call) but the `tools:` will be supported by the new `Langchain::Assistant` class: https://github.com/andreibondarev/langchainrb/pull/422

PR #288, opened 2023-08-11
rename tool_name and tool_description to name and description

diego-suarez:
I think its a bit redundant to have the `tool_` prefix on the tools attributes


andreibondarev:
@diego-suarez Failing specs

PR #289, opened 2023-08-12
Run CI on push on the main branch

santib:
We are currently not running the Tests when something gets merged to `main`


.github/workflows/ci.yml
```diff
@@ -6,7 +6,7 @@ on:
       - "*"
   push:
     branches:
-      - master
```
andreibondarev, line 9: Oh man, nice find! 😆

PR #297, opened 2023-08-21
Add k to pinecone's "ask" functionality

asmorris:
We were running into an issue when asking a question of OpenAI using pinecone, where there were occasions where the context provided by the `ask` method was over the 4096 tokens, and so was exceeding the default OpenAI context window for question asking. This change would allow a user to add in an additional parameter `k` to the ask method, which could be subsequently passed to `similarity_search` to allow the user to have the amount of relevant results they wish, instead of defaulting to the 4 that is with `similarity_search` currently.

### DIFF
--- a/lib/langchain/vectorsearch/pinecone.rb
+++ b/lib/langchain/vectorsearch/pinecone.rb
@@ -142,7 +142,7 @@ module Langchain::Vectorsearch
     # @yield [String] Stream responses back one String at a time
     # @return [String] The answer to the question
     def ask(question:, namespace: "", filter: nil, k: 4, &block)
-      search_results = similarity_search(query: question, namespace: namespace, filter: filter)
+      search_results = similarity_search(query: question, namespace: namespace, filter: filter, k: k)
 
       context = search_results.map do |result|
         result.dig("metadata").to_s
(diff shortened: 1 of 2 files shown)


lib/langchain/vectorsearch/pinecone.rb
```diff
@@ -141,7 +141,7 @@ def similarity_search_by_vector(embedding:, k: 4, namespace: "", filter: nil)
     # @param filter [String] The filter to use
     # @yield [String] Stream responses back one String at a time
     # @return [String] The answer to the question
-    def ask(question:, namespace: "", filter: nil, &block)
+    def ask(question:, namespace: "", filter: nil, k: 4, &block)
       search_results = similarity_search(query: question, namespace: namespace, filter: filter)
```
andreibondarev, line 145: I think you meant to also pass the `k:` param to the `similarity_search(..., k: k)`

asmorris: 🤦  yes - absolutely did mean to do that. long day sorry.

andreibondarev (approved):

andreibondarev:
@asmorris I wouldn't be opposed to adding the same param to all other vector search DBs if you feel like coding it up!

asmorris:
> I wouldn't be opposed to adding the same param to all other vector search DBs if you feel like coding it up!

I can absolutely do that! Great idea

PR #298, opened 2023-08-22
Add k to all asks

asmorris:
### Context
This PR is the logical extension of https://github.com/andreibondarev/langchainrb/pull/297 - we were running into limits with the `ask` method where our context was larger than the 4096 token window available for our OpenAI model, and so we needed to slim it down by reducing the amount of results we received. However there was no `k` available on the method, so this gives all those methods that functionality.

### DIFF
--- a/lib/langchain/vectorsearch/pinecone.rb
+++ b/lib/langchain/vectorsearch/pinecone.rb
@@ -138,6 +138,7 @@ module Langchain::Vectorsearch
     # Ask a question and return the answer
     # @param question [String] The question to ask
     # @param namespace [String] The namespace to search in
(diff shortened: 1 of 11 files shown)


lib/langchain/vectorsearch/pinecone.rb
```diff
@@ -138,6 +138,7 @@ def similarity_search_by_vector(embedding:, k: 4, namespace: "", filter: nil)
     # Ask a question and return the answer
     # @param question [String] The question to ask
     # @param namespace [String] The namespace to search in
+    # @param k [Integer] The number of results to have in context
```
asmorris, line 141: this change is because I missed adding this in the comment to my previous PR.

PR #300, opened 2023-08-23
add namespace to add_data method

asmorris:
### Context
Adds namespace as a parameter to `add_data` method so that a user can specify which namespace they want their uploaded data to end up associated with.

### DIFF
--- a/lib/langchain/vectorsearch/base.rb
+++ b/lib/langchain/vectorsearch/base.rb
@@ -160,7 +160,7 @@ module Langchain::Vectorsearch
       prompt_template.format(question: question)
     end
 
-    def add_data(paths:)
+    def add_data(paths:, namespace: "")
       raise ArgumentError, "Paths must be provided" if Array(paths).empty?
 
       texts = Array(paths)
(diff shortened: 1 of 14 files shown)


andreibondarev:
@asmorris Thank you for another PR! I'm just wondering since the namespaces are currently only a Pinecone concept -- should we just overwrite `add_data()` method on the Pinecone class itself? I know it's a little non-DRY but it might be a cleaner approach for now.

asmorris:
> @asmorris Thank you for another PR! I'm just wondering since the namespaces are currently only a Pinecone concept -- should we just overwrite add_data() method on the Pinecone class itself? I know it's a little non-DRY but it might be a cleaner approach for now.

that's a great idea - i think having a bit less DRY-ness is worth the trade off vs the confusion of functionality that doesn't make sense for the other classes. I will modify this!

andreibondarev:
> > @asmorris Thank you for another PR! I'm just wondering since the namespaces are currently only a Pinecone concept -- should we just overwrite add_data() method on the Pinecone class itself? I know it's a little non-DRY but it might be a cleaner approach for now.
> 
> that's a great idea - i think having a bit less DRY-ness is worth the trade off vs the confusion of functionality that doesn't make sense for the other classes. I will modify this!

Thank you, and let's 🚢 it!

PR #310, opened 2023-09-05
Avoid duplications when requiring gem dependencies

### LINKED ISSUE #306: Combine the functionality into a single method call.
Example: https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/llm/openai.rb#L29-L30

This type of code is repeated in many many places. Instead of having 2 lines, we should just be able to call a single method that does both `depends_on()` and `require()`.

Maybe `depends_on` itself does the require as well?

dferrazm:
DRY `depends_on` directives to require the gem code automatically.

As an extra: I'm scoping the `::LoadError` raised by the dependency helper under the same namespace (like the `VersionError`, for consistency).

Fixes #306


lib/langchain/dependency_helper.rb
```diff
@@ -26,8 +28,13 @@ def depends_on(gem_name)
         raise VersionError, "The #{gem_name} gem is installed, but version #{gem_requirement} is required. You have #{gem_version}."
       end
 
+      lib_name = gem_name if req == true
+      lib_name = req if req.is_a?(String)
```
dferrazm, line 32: Added the explict `req == true` check so the order doesn't affect the result. Otherwise, if we had only:

```rb
lib_name = gem_name if req
lib_name = req if req.is_a?(String)
```

... it would also work, but order would matter.

andreibondarev (approved):
@dferrazm Really nicely done!

PR #311, opened 2023-09-06
Handle OpenAI API errors

alchaplinsky:
## Motivation
Currently, if OpenAI API returns an error langchainrb's  `complete` method just returns `nil` and `chat` raises an error. Making this behavior consistent across all methods so that developers actually see an error. 

## Changes
- Add `with_api_error_handling` method to OpenAI LLM that checks for errors in API responses
- Add test cases


andreibondarev (approved):

andreibondarev:
Nice!

PR #312, opened 2023-09-06
Update OpenAI completion to use chat API endpoint

alchaplinsky:
## Motivation
Completions API is deprecated. Models used for text completions are going to be shut down on 4 Jan 2024  https://platform.openai.com/docs/deprecations/2023-07-06-gpt-and-embeddings 

OpenAI recommends using chat endpoint instead.

## Changes
- Switch LLM's `complete ` method to using `chat` endpoint (using chat method on OpenAI client) and chat_completion_model by default.
- Update complete tests accordingly

## Notes
Depends on https://github.com/andreibondarev/langchainrb/pull/311

### DIFF
--- a/lib/langchain/llm/openai.rb
+++ b/lib/langchain/llm/openai.rb
@@ -12,10 +12,19 @@ module Langchain::LLM
   class OpenAI < Base
     DEFAULTS = {
       temperature: 0.0,
+      completion_model_name: "gpt-3.5-turbo",
       chat_completion_model_name: "gpt-3.5-turbo",
       embeddings_model_name: "text-embedding-ada-002",
       dimension: 1536
     }.freeze
+
+    LEGACY_COMPLETION_MODELS = %w[
+      ada
+      babbage
+      curie
+      davinci
+    ].freeze
+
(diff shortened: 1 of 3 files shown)


d96litz:
Nice, thanks for this! This will be a breaking change right? People who configured the model to "text-davinci-xxx" will get an error. I don't know if there are any implications to this. Just wanted to mention.

alchaplinsky:
@d96litz yeah, it'll be a breaking change. I don't know if it makes sense to temporarily support a legacy endpoint and just deprecate it with a warning. I'd just add a note about breaking change to release notes.

alchaplinsky:
Okay, supporting legacy models for now. 
Using `completions` API in case if a model is explicitly set to one of the legacy models via LLM `default_options` or direct `model` option on `complete` method.

**passing a  legacy model to `default_options` into constructor:**

bin:console 2023-09-10 19-29-08

or

**passing a legacy model via `model` parameter**

bin:console 2023-09-10 19-26-44

cc @d96litz @andreibondarev

andreibondarev (approved):

PR #338, opened 2023-10-03
👷‍♂️ Conversations: Allow OpenAI LLM to return multiple choices [Part 3]

alchaplinsky:
## Motivation
Previously we had `response.dig("choices", 0, "message")` hardcoded in our `LLM::OpenAI#chat` method. Which means that we don't support multiple-choice responses and always return only the first choice. The number of choices is controlled by `n` parameter that can be passed to `initializer` as default options or to the `chat` method.

## Changes
- Change the `chat` method to return either a string (in case of a single choice) or an array of strings (in case of multiple choices)
- If a block is given to handle streaming response the entire hash is yielded to the block as it contains an `index` of a choice as well as `delta`.
- Update specs to account for the default `n` param that is sent to open ai API.

### DIFF
--- a/lib/langchain/llm/openai.rb
+++ b/lib/langchain/llm/openai.rb
@@ -11,6 +11,7 @@ module Langchain::LLM
   #
   class OpenAI < Base
     DEFAULTS = {
+      n: 1,
       temperature: 0.0,
       completion_model_name: "gpt-3.5-turbo",
       chat_completion_model_name: "gpt-3.5-turbo",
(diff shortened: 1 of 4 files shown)


lib/langchain/llm/openai.rb
```diff
@@ -229,5 +232,10 @@ def with_api_error_handling
     def validate_max_tokens(messages, model)
       LENGTH_VALIDATOR.validate_max_tokens!(messages, model)
     end
+
+    def extract_response(response)
+      results = response.dig("choices").map { |choice| choice.dig("message", "content") }
+      (results.size == 1) ? results.first : results
```
andreibondarev, line 238: I think it's a little odd that this method returns either a string or an array (different data structures). But maybe we can change this when we introduce `Langchain::LLM::Response`

alchaplinsky: Yeah, that's what I was thinking

andreibondarev (approved):

PR #340, opened 2023-10-06
Introduce Langchain::LLM::Response object

alchaplinsky:
## Motivation
We need to have unified LLM response object that would include all the useful information that is contained in the LLM API response and yet will be consistent across different LLMs.

## Changes
- Introduce `Langchain::LLM::Response::OpenAI` and `Langchain::LLM::Response::GooglePalm` result classes
- Update `embed`, `complete` and `chat` methods on OpenAI and Google's LLM to return result object
- Update specs

## Usage
```ruby
result = llm.chat(prompt: 'Hello')

result.model # => gpt-3.5-turbo
result.type # => chat.completion
result.completions # => [{role: '...', content: '...'}]
result.value # => completion text
result.prompt_tokens # => 1
result.completion_tokens # => 5
result.total_tokens # => 6
```


lib/langchain/llm/responses/google_palm.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+module Langchain::LLM::Responses::GooglePalm
```
andreibondarev, line 3: Why not just do a classic class inheritance structure, like the following (oversimplified) below?

```ruby
class Langchain::LLM::Response
  def initialize(raw_response)
     @raw_response = raw_response
  end

  def completion_text
```

```ruby
class Langchain::LLM::GooglePalmResponse < Langchain::LLM::Response
  def completion_text
     completions.first.dig("output")
  end

  def completions
    raw_response.dig("candidates")
  end
```

alchaplinsky: Well, that's one way to do it. In this case, all the LLMs will return instances of different classes. I was trying to achieve a consistent result.

alchaplinsky:
@andreibondarev Ok, tried this approach. And I actually like it better, so updated.

.gitignore
```diff
@@ -14,3 +14,4 @@
 .env
 .DS_Store
 .idea
+test.ann
```
andreibondarev, line 17: Nice find!

andreibondarev:
Shouldn't this also respond to `.model()`?
```ruby
irb(main):004:0> llm = Langchain::LLM::GooglePalm.new(api_key: ENV['GOOGLE_PALM_API_KEY'])
=>
#<Langchain::LLM::GooglePalm:0x0000000110b0ab10
...
irb(main):005:0> llm.model
(irb):5:in `<main>': undefined method `model' for #<Langchain::LLM::GooglePalm:0x0000000110b0ab10 @client=#<GooglePalmApi::Client:0x0000000110b0a2a0 @api_key="AIzaSyBRzMe0CRTkgIYctVPTy9eepY2
```

lib/langchain/llm/google_palm.rb
```diff
@@ -34,21 +34,20 @@ def initialize(api_key:, default_options: {})
...
+    # @return [Langchain::LLM::Response::GooglePalm] LLM response object with embedding as a value
     #
     def embed(text:)
-      response = client.embed(
-        text: text
-      )
-      response.dig("embedding", "value")
+      response = client.embed(text: text)
+
+      Langchain::LLM::Response::GooglePalm.new response, model: @defaults[:embeddings_model_name], type: "embedding"
```
andreibondarev, line 42: Why do we need to set the `type:` here? Doesn't the caller know what they should be getting back? Isn't up to them to check `.value` or `.completions` or `.embedding`?

lib/langchain/llm/response/openai.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def prompt_tokens
+      raw_response.dig("usage", "prompt_tokens")
+    end
+
+    def completion_tokens
+      raw_response.dig("usage", "completion_tokens")
+    end
+
+    def total_tokens
```
andreibondarev, line 37: Thoughts on calling these: `total_tokens_used`, `completion_tokens_used`, `prompt_tokens_used`? So it's even more clear.

lib/langchain/llm/response/openai.rb
```diff
@@ -0,0 +1,47 @@
...
+
+module Langchain::LLM::Response
+  class OpenAI < Base
+    attr_reader :raw_response
+
+    def model
+      raw_response["model"]
+    end
+
+    def type
```
andreibondarev, line 11: I'm wondering if we should just fully simplify this and get this of the `value()` method and the needed `type()`.

lib/langchain/llm/google_palm.rb
alchaplinsky: If caller does a single call and gets a response, they probably should. But what if you fire a bunch of requests to LLM(s) and then just push responses to the queue to deal with them later? I think that response should contain information about what type of response it is.

lib/langchain/llm/response/openai.rb
alchaplinsky: seems like just more characters to type to me 😄

lib/langchain/llm/google_palm.rb
andreibondarev: I'd probably suggest that we don't solve problems we don't currently have. YAGNI.

alchaplinsky:
> Shouldn't this also respond to `.model()`?
> 
> ```ruby
> irb(main):004:0> llm = Langchain::LLM::GooglePalm.new(api_key: ENV['GOOGLE_PALM_API_KEY'])
> =>
> #<Langchain::LLM::GooglePalm:0x0000000110b0ab10
> ...
> irb(main):005:0> llm.model
> (irb):5:in `<main>': undefined method `model' for #<Langchain::LLM::GooglePalm:0x0000000110b0ab10 @client=#<GooglePalmApi::Client:0x0000000110b0a2a0 @api_key="AIzaSyBRzMe0CRTkgIYctVPTy9eepY2
> ```
> 
> and the rest of the methods too, like `.completions`, `.embedding`, `.value`, etc.

What do you mean? This is an LLM instance, what would it return as a value?

lib/langchain/llm/google_palm.rb
alchaplinsky: Well, I kinda already need to know the type within the response itself in order to properly handle data returned from the LLM.

andreibondarev:
> > Shouldn't this also respond to `.model()`?
> > ```ruby
> > irb(main):004:0> llm = Langchain::LLM::GooglePalm.new(api_key: ENV['GOOGLE_PALM_API_KEY'])
> > =>
> > #<Langchain::LLM::GooglePalm:0x0000000110b0ab10
> > ...
> > irb(main):005:0> llm.model
> > (irb):5:in `<main>': undefined method `model' for #<Langchain::LLM::GooglePalm:0x0000000110b0ab10 @client=#<GooglePalmApi::Client:0x0000000110b0a2a0 @api_key="AIzaSyBRzMe0CRTkgIYctVPTy9eepY2
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and the rest of the methods too, like `.completions`, `.embedding`, `.value`, etc.
> 
> What do you mean? This is an LLM instance, what would it return as a value?

@alchaplinsky D'oh! My bad -- needed to call `llm2.complete()` or other methods obviously.

andreibondarev:
@alchaplinsky Lots of comments but great work thus far!

lib/langchain/llm/response/base.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+module Langchain::LLM::Response
+  class Base
```
andreibondarev, line 4: Thoughts on renaming the class structure to:
```
Langchain::LLM::Response # parent
Langchain::LLM::OpenAIResponse # child
Langchain::LLM::GooglePalmResponse # child
```
?

alchaplinsky: Would be cool to have less namespaces, but I'm concerned about file structure.. this will result in a bunch of file living in the `llm` directory which are not necessarily LLMs but also responses which is a bit messy. Usually ruby/rails approach is to group (namespace) classes by their types.

andreibondarev: Can't we put it all still in the `langchain/llm/response/` directory, independent from what the classes are called?

lib/langchain/llm/response/openai.rb
andreibondarev: I was wondering if it brings more clarity?

lib/langchain/llm/google_palm.rb
andreibondarev: In [this comment](https://github.com/andreibondarev/langchainrb/pull/340#discussion_r1349752543) I was suggesting simplifying it further and getting rid of the whole `type` and `value` methods.

lib/langchain/llm/response/base.rb
alchaplinsky: maybe... need to look at zeitwerk's documentation

lib/langchain/llm/google_palm.rb
alchaplinsky: If you have some specific interface for this object in mind, please share. I don't really know what you mean by "simplifying through getting rid of value method".

lib/langchain/llm/response/base.rb
alchaplinsky: found a way to change this

lib/langchain/llm/response/openai.rb
alchaplinsky: I'm not sure if `used` is bringing more clarity.. It is usage from OpenAI API limits perspective, but in context of Response object it is just counts of tokens in prompt and completion. 

I'd keep it short and named according to the class context, rather than overall paradigm of API's tokens usage/limits.

alchaplinsky:
hey @andreibondarev, would you mind giving it another view?

andreibondarev:
> hey @andreibondarev, would you mind giving it another view?

Yep, I'm actively adding more code to this, will be pushing today.

andreibondarev:
@alchaplinsky I'll merge this for now, we can keep enhancing it in follow-up PRs.

Issue #320, opened 2023-09-18
Conversation message history is broken


alchaplinsky:
## Details
Passing an array of messages as hashes to a conversation leads to errors when trying to send a message
```ruby 
messages = [
  {role: "user", message: "..."},
  {role: "ai", message: "..."},
]
conversation = Langchain::Conversation.new(llm: llm, messages: messages)
conversation.message('hello')

NoMethodError (undefined method `type' for {}:Hash
```

drale2k:
I think you can't pass hashes directly but need to fill the array with `HumanMessage` and `AIMessage` instances. 

```ruby
human_message = HumanMessage.new("hello")
ai_message = AIMessage.new("hi")

message = [human_message, ai_message]
```

alchaplinsky:
It was possible before this one https://github.com/andreibondarev/langchainrb/pull/276 got merged. 

I thought the idea was that `Langchain::Conversation` wraps messages internally with `HumanMessage` and `AIMessage`. But it appears that it also returns messages as instances of those classes and expects `examples` and `messages` to be that.

Anyway, `Langchain::Conversation`  constructor SHOULD NOT just silently assume that passed messages are of correct kind. And then just raise an error when you try to send a message. Conversation should either smartly wrap passed `messages` with these classes or raise an error in `initializer` that messages are of incorrect kind.

PR #341, opened 2023-10-07
Improve Langchain::Conversation: allow passing array of hashes as messages and examples

alchaplinsky:
## Motivation
If you have a history of messages in plain hash format it is tedious to wrap them into `Langchain::Conversation::Prompt`/`Langchain::Conversation::Response`  classes before passing them to `Langchain::Conversation`. Conversation class can be more intelligent and do that for the user.

## Changes
Allow Conversation to accept messages objects and hashes as parameters on initializer

## Resolves
Issue https://github.com/andreibondarev/langchainrb/issues/320

### DIFF
--- a/lib/langchain/conversation.rb
+++ b/lib/langchain/conversation.rb
@@ -22,15 +22,14 @@ module Langchain
     # Intialize Conversation with a LLM
     #
     # @param llm [Object] The LLM to use for the conversation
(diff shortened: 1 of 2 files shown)


andreibondarev:
I've deprecated the `Langchain::Conversation` class in an upcoming PR.

PR #362, opened 2023-10-21
Change zeitwerk version to ~> 2.5 from explicit = version

ci:
👋 @santib @andreibondarev - noticed while adding the gem with a Rails 7.0 app, that it auto-resolved an earlier version (0.6.11) - i.e. related to https://github.com/andreibondarev/langchainrb/pull/278

railties on the 7.0 version uses ~> 2.5 so it doesn't allow latest langchainrb and Rails 7.0 to co-exist; there doesn't seem to be an explicit reason for 2.6.11 so allowing an earlier version here.

https://github.com/rails/rails/blob/v7.0.8/railties/railties.gemspec#L46

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -16,7 +16,7 @@ PATH
       json-schema (~> 4.0.0)
       pragmatic_segmenter (~> 0.3.0)
       tiktoken_ruby (~> 0.0.5)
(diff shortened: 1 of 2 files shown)


andreibondarev:
@ci Nice! Thank you for the contribution! (Also -- nice Github handle ;)

PR #376, opened 2023-11-02
Add support for OpenAI via Azure

duncantmiller:
Hi Andrei,

I've made an attempt at adding support for the Microsoft [Azure OpenAI Service](https://learn.microsoft.com/en-us/azure/ai-services/openai/) API. It uses the [ruby-openai ](https://github.com/alexrudall/ruby-openai) gem and there is some basic documentation there about using Azure. The main difference is that with Azure you deploy multiple models, one for embeddings and one for chat.  So you need to specify the `uri_base` parameter whenever you call the `OpenAI::Client` with the url of either your embed model deployment (ada-2) or your chat model deployment (gpt-35) based on the proper context.

I decided to just have the Azure class inherit from the OpenAI class since they are so similar and just update a few of the methods. I'm not sure if this is the best approach, or if you think I should try something else, perhaps keeping the classes separate but including the shared methods in a module? I also wasn't totally sure on the how to best approach specs so included some azure specs for the new methods but I also copied in a lot of redundant specs to the spec/langchain/llm/azure_spec.rb file that would call methods in the OpenAI class and are already tested in spec/langchain/llm/openai_spec.rb.

Also I had to update to version 5.2 of the ruby-openai gem in development in order to use Azure.

Please review and let me know if you have any thoughts or changes I could make. Thanks!

### DIFF
--- a/lib/langchain/llm/azure.rb
+++ b/lib/langchain/llm/azure.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Langchain::LLM
-  # LLM interface for OpenAI APIs: https://platform.openai.com/overview
+  # LLM interface for Azure OpenAI Service APIs: https://learn.microsoft.com/en-us/azure/ai-services/openai/
   #
   # Gem requirements:
   #    gem "ruby-openai", "~> 4.0.0"
--- a/lib/langchain/llm/azure.rb
+++ b/lib/langchain/llm/azure.rb
@@ -1,0 +1,140 @@
+# frozen_string_literal: true
+
+module Langchain::LLM
+  # LLM interface for OpenAI APIs: https://platform.openai.com/overview
+  #
+  # Gem requirements:
+  #    gem "ruby-openai", "~> 4.0.0"
+  #
+  # Usage:
+  #    openai = Langchain::LLM::Azure.new(api_key:, llm_options: {}, embedding_deployment_url: chat_deployment_url:)
+  #
+  class Azure < OpenAI
+    attr_reader :embed_client
+    attr_reader :chat_client
+
+    def initialize(
+      api_key:,
+      llm_options: {},
+      default_options: {},
+      embedding_deployment_url: nil,
+      chat_deployment_url: nil
+    )
+      depends_on "ruby-openai", req: "openai"
+
+      @embed_client = ::OpenAI::Client.new(
+        access_token: api_key,
+        uri_base: embedding_deployment_url,
+        **llm_options
+      )
+      @chat_client = ::OpenAI::Client.new(
+        access_token: api_key,
+        uri_base: chat_deployment_url,
+        **llm_options
+      )
+      @defaults = DEFAULTS.merge(default_options)
+    end
+
+    #
+    # Generate an embedding for a given text
+    #
+    # @param text [String] The text to generate an embedding for
+    # @param params extra parameters passed to OpenAI::Client#embeddings
+    # @return [Langchain::LLM::OpenAIResponse] Response object
+    #
+    def embed(text:, **params)
+      parameters = {model: @defaults[:embeddings_model_name], input: text}
+
+      validate_max_tokens(text, parameters[:model])
+
+      response = with_api_error_handling do
+        embed_client.embeddings(parameters: parameters.merge(params))
+      end
+
+      Langchain::LLM::OpenAIResponse.new(response)
+    end
+
+    #
+    # Generate a completion for a given prompt
+    #
+    # @param prompt [String] The prompt to generate a completion for
+    # @param params  extra parameters passed to OpenAI::Client#complete
+    # @return [Langchain::LLM::Response::OpenaAI] Response object
+    #
+    def complete(prompt:, **params)
+      parameters = compose_parameters @defaults[:completion_model_name], params
+
+      parameters[:messages] = compose_chat_messages(prompt: prompt)
+      parameters[:max_tokens] = validate_max_tokens(parameters[:messages], parameters[:model])
+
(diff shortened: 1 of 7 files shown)


lib/langchain/llm/azure.rb
```diff
@@ -0,0 +1,139 @@
+# frozen_string_literal: true
+
+module Langchain::LLM
+  # LLM interface for Azure OpenAI Service APIs: https://learn.microsoft.com/en-us/azure/ai-services/openai/
+  #
+  # Gem requirements:
+  #    gem "ruby-openai", "~> 4.0.0"
+  #
+  # Usage:
+  #    openai = Langchain::LLM::Azure.new(api_key:, llm_options: {}, embedding_deployment_url: chat_deployment_url:)
```
andreibondarev, line 10: Since it's just the last part of the URL that changes specifying the `model_name`, e.g.:
```
https://custom-domain.openai.azure.com/openai/deployments/gpt-35-turbo
https://custom-domain.openai.azure.com/openai/deployments/ada-2
```

Any reason why we wouldn't treat them like model names and adhere to the interface that all other LLMs follow, like such:
```ruby
    DEFAULTS = {
      chat_completion_model_name: "gpt-35-turbo",
      embeddings_model_name: "ada-2"
      ...
    }.freeze
```

And then, when initializing the client, we'd just append the model name:
```ruby
uri_base: (AZURE_URL + model_name)
```

The idea is that there are defaults but you can override them by specifying the model names you'd like to use.

andreibondarev (commented):
@duncantmiller Thank you for this PR!

andreibondarev:
@duncantmiller Are there any additional attributes that come back from the Azure OpenAI that the [response class](https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/llm/response/openai_response.rb) could be extended with?

duncantmiller:
> @duncantmiller Are there any additional attributes that come back from the Azure OpenAI that the [response class](https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/llm/response/openai_response.rb) could be extended with?

It appears that the raw response is the same from both APIs. Here is the Azure API response:

````
#<Langchain::LLM::OpenAIResponse:0x000000010c340848
 @model=nil,
 @raw_response=
  {"id"=>"chatcmpl-8GV2shoXGnRT0JhV7l5Id4Wjj9zpn",
   "object"=>"chat.completion",
   "created"=>1698943278,
   "model"=>"gpt-35-turbo-16k",
   "choices"=>
    [{"index"=>0,
      "finish_reason"=>"stop",
      "message"=>
       {"role"=>"assistant",
        "content"=>
         "I'm sorry, but I cannot provide an answer without more information. Could you please provide more context or clarify your question?"}}],
   "usage"=>{"prompt_tokens"=>17, "completion_tokens"=>25, "total_tokens"=>42}}>
````

And the OpenAI API response

````
#<Langchain::LLM::OpenAIResponse:0x00000001071cf608
 @model=nil,
 @raw_response=
  {"id"=>"chatcmpl-8GV5xEZgNsfI2XYkBI2vw0m3eUnQ9",
   "object"=>"chat.completion",
   "created"=>1698943469,
   "model"=>"gpt-3.5-turbo-0613",
   "choices"=>
    [{"index"=>0,
      "message"=>
       {"role"=>"assistant",
        "content"=>
         "I'm sorry, but I cannot provide an answer without more information. Could you please provide more context or clarify your question?"},
      "finish_reason"=>"stop"}],
   "usage"=>{"prompt_tokens"=>17, "completion_tokens"=>25, "total_tokens"=>42}}>
````

lib/langchain/llm/azure.rb
duncantmiller: I like this idea and I think it would work but I wonder if it might be too limiting. The last part of the url is actually the `deployment_name`, when you create your deployment on Azure you pick a model and then create a deployment name. I used the model names in the examples and we could force that convention.

However it is possible to have multiple Azure deployments with the same model, like for example I'm not sure if there is a reason for this but someone might want to have separate development and production deployments with names like `gpt-3-5-development` and `gpt-3-5-production`. 

They also allow you to customize each model and train it on your own dataset or fine tune it, though that does end up resulting in a new model with a specific name (from the [API docs](https://learn.microsoft.com/en-us/azure/ai-services/openai/tutorials/fine-tune?tabs=command-line#deploy-fine-tuned-model)): 

````
fine_tuned_model 

Retrieve this value from your fine-tuning job results in the previous step. 
It will look like gpt-35-turbo-0613.ft-b044a9d3cf9c4228b5d393567f693b83. 
You will need to add that value to the deploy_data json
````

It also appears that they have models with the same model name but different versions, for example `text-embedding-ada-002` is the same model name for both version 1 and 2 which [are different](https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models#embeddings). See screenshot below from my Azure dashboard model options:

Screen Shot 2023-11-02 at 10 33 28 AM

andreibondarev: @duncantmiller Thank you for this detailed explanation, and I agree with your original approach now!

andreibondarev:
@duncantmiller Thank you, thank you! I'll be cutting a new gem version shortly!

lib/langchain/llm/azure.rb
duncantmiller: I was just playing around with my Azure account a bit and realized I can create multiple services with different custom domain names (the first part of the url) so that could solve the development / production separation. So for example you could have:

`https://foo-development.openai.azure.com/openai/deployments/gpt-35-turbo`

and

`https://foo-production.openai.azure.com/openai/deployments/gpt-35-turbo`

PR #384, opened 2023-11-09
[WIP] Add AWS Bedrock LLM

fdaugs:
## Changes Overview

This PR adds an new llm called `AwsBedrock` for these models running on AWS Bedrock:
- ai21.j2-mid-v1
- ai21.j2-ultra-v1
- anthropic.claude-instant-v1
- anthropic.claude-v1
- anthropic.claude-v2
- cohere.command-text-v14
- amazon.titan-embed-text-v1

## TODO

- [ ] Add token validation?
- [ ] Documentation?
- [ ] Examples?
- [x] Add embedding API with Amazons Titan Embeddings G1 model
- [x] Add specs


fdaugs:
Hey @andreibondarev. I have basic implementation for the bedrock api which "works on my machine"<sup>TM</sup>. The complex part is, that the bedrock api is basically just an "access point" to use the models by ai21, anthropic and cohere. As a result, each model has a different api with different parameters and responses.

I would be interested to know what you think a good common API might be (i.e. the DEFAULTS hash). Do you see any potential problems in returning the existing response classes depending on the chosen model?

andreibondarev:
@fdaugs Reviewing it now...

andreibondarev:
@fdaugs Could you please mention how to properly initialize the `Langchain::LLM::AwsBedrock ` client? I'll go digging but you'd save me a lot of time.

fdaugs:
Yeah sure! The simplest form is actually no arguments at all:
```
bedrock = Langchain::LLM::AwsBedrock.new
bedrock.complete(prompt: "Hello World").completion
```
This will use the `anthropic.claude-v2` model.

But you can also pass in additional arguments:
```
bedrock = Langchain::LLM::AwsBedrock.new(completion_model: "anthropic.claude-v1", aws_client_options: {region: 'us-east-1'}, default_options:{temperature: 0.7})
```
For `aws_client_options` all option, that are described here: `https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/BedrockRuntime/Client.html#initialize-instance_method`. 
If you have already authenticated with the aws-cli and configured the correct IAM policies for your profile to access the Bedrock service, you should not need to pass in any additional parameters (apart from the region if you are accessing models in a region other than your default region). All configuration options for using the aws client are automatically read from these files: `~/.aws/credentials` `~/.aws/config` or environment variables.

andreibondarev:
@fdaugs Nice! I was able to test out both the complete() and embed() methods. Working great thus far!

andreibondarev:
> Hey @andreibondarev. I have basic implementation for the bedrock api which "works on my machine"TM. The complex part is, that the bedrock api is basically just an "access point" to use the models by ai21, anthropic and cohere. As a result, each model has a different api with different parameters and responses.
> 
> I would be interested to know what you think a good common API might be (i.e. the DEFAULTS hash). Do you see any potential problems in returning the existing response classes depending on the chosen model?

I've debated making some sort of a split into `Langchain::LLM::Provider` and `Langchain::LLM::Model` because the single `Langchain::LLM::*` is conflating the two. For example Claude-2 will be available via `Langchain::LLM::Anthropic` and now `Langchain:LLM::AwsBedrock`. I'm not sold that this change needs to be immediately thought. What do you think?

fdaugs:
Okay and the `Langchain::LLM::Model` would just be a wrapper of the input options and Response? This cloud be potentially harder to maintain if the "real" API e.g. from AI21 ever differs from the one provided by AWS. I think this cloud be already the case. For the example the offical api docs of the jurassic model contains configuration options like "minTokes" and "numResults" that aren't mentioned in the AWS docs.

The [implementation of langchain](https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/llms/bedrock.py) also implements Bedrock as a single "LLM".

andreibondarev:
> Okay and the `Langchain::LLM::Model` would just be a wrapper of the input options and Response? This cloud be potentially harder to maintain if the "real" API e.g. from AI21 ever differs from the one provided by AWS. I think this cloud be already the case. For the example the offical api docs of the jurassic model contains configuration options like "minTokes" and "numResults" that aren't mentioned in the AWS docs. Not to mention the fact that different providers already offer different models.
> 
> The [implementation of langchain](https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/llms/bedrock.py) also implements Bedrock as a single "LLM".

Yeah, different LLM providers can definitely implement different options, I don't think it's a guarantee. Alright -- let's keep it as is like you're suggesting!

langchain.gemspec
```diff
@@ -45,6 +45,7 @@ Gem::Specification.new do |spec|
   # optional dependencies
   spec.add_development_dependency "ai21", "~> 0.2.1"
   spec.add_development_dependency "anthropic", "~> 0.1.0"
+  spec.add_development_dependency "aws-sdk-bedrockruntime", "~> 1.1"
```
andreibondarev, line 48: @fdaugs I added this so it's easy to do local dev. I hope you don't mind!

andreibondarev (approved):

fdaugs:
@andreibondarev 
Wait a moment before merging. I should probably remove the unused validation method...

andreibondarev:
@fdaugs Also the specs are failing.

fdaugs:
Seems link the aws client call is not properly mocked. I will have a look.

fdaugs:
@andreibondarev This should have solved the issue with the specs.

andreibondarev:
@fdaugs Safe to merge from your perspective?

fdaugs:
If the last commit had fixed the specs yes ;) I don't quite get, why the AWS_Region env is still missing. When commenting out the default region in my aws config I got the same error as in the ci and stubbing the env made the specs pass. I cloud mock the whole client but then configuring the client would't be covered by the specs anymore.

andreibondarev:
> If the last commit had fixed the specs yes ;) I don't quite get, why the AWS_Region env is still missing. When commenting out the default region in my aws config I got the same error as in the ci and stubbing the env made the specs pass. I cloud mock the whole client but then configuring the client would't be covered by the specs anymore.

So if this is the error:
```
       No region was provided. Configure the `:region` option or export the region name to ENV['AWS_REGION']
```
I'm wondering if we can just add that env var to the [ci.yml](https://github.com/andreibondarev/langchainrb/blob/main/.github/workflows/ci.yml)?

andreibondarev:
@fdaugs Looks like that did the trick! ✅ Thank you so much for this contribution! 🙏🏻

PR #387, opened 2023-11-11
Handle conversation message when block is passed (issue #364)

mathieuripert:
Suggested fix for issue #364

It changes `Conversation#message` method to handle nil responses from OpenAI `chat` when a streaming block is used. Now, it accumulates response chunks from the block to reconstruct the final response to append it the chat history.

I also modified the streaming block in `openai.rb` to yield the "content" directly from the response chunk, instead of the entire first "choice". This is debatable here, the alternative would be get the content in conversation class, but this seems to be LLM specific.

## EDIT:
Followed comments from @alchaplinsky and implemented the response chunks accumulation at lower level in `Langchain::LLM::OpenAI` classs. `Conversation` class stays intact. 
This change ensures that openai llm chat alway return a response for bot streaming and non streaming.

Example
```ruby
response = llm.chat(prompt: "hello") do |chunk|
  puts chunk
end
```
this code will return 
```
{"index"=>0, "delta"=>{"role"=>"assistant", "content"=>""}, "finish_reason"=>nil}
{"index"=>0, "delta"=>{"content"=>"Hello"}, "finish_reason"=>nil}
{"index"=>0, "delta"=>{"content"=>"!"}, "finish_reason"=>nil}
{"index"=>0, "delta"=>{"content"=>" How"}, "finish_reason"=>nil}
{"index"=>0, "delta"=>{"content"=>" can"}, "finish_reason"=>nil}
{"index"=>0, "delta"=>{"content"=>" I"}, "finish_reason"=>nil}
{"index"=>0, "delta"=>{"content"=>" assist"}, "finish_reason"=>nil}
{"index"=>0, "delta"=>{"content"=>" you"}, "finish_reason"=>nil}
{"index"=>0, "delta"=>{"content"=>" today"}, "finish_reason"=>nil}
{"index"=>0, "delta"=>{"content"=>"?"}, "finish_reason"=>nil}
{"index"=>0, "delta"=>{}, "finish_reason"=>"stop"}
=> 
#<Langchain::LLM::OpenAIResponse:0x000000010733ba78
 @model=nil,
 @raw_response=
  {"id"=>"chatcmpl-8KMFBz0dfSVKbB5mzUgguZhpODkBH",
   "object"=>"chat.completion.chunk",
   "created"=>1699862757,
   "model"=>"gpt-3.5-turbo-0613",
   "choices"=>[{"message"=>{"role"=>"assistant", "content"=>"Hello! How can I assist you today?"}}]}>
````

### CODE REVIEW

**Comment by andreibondarev** [comment]
@mathieuripert Lmk when you'd like me to review this one. Thank you for deciding to tackle this one!

**Comment by mathieuripert** [comment]
> @mathieuripert Lmk when you'd like me to review this one. Thank you for deciding to tackle this one!

it should be good for review. Thanks for looking at it @andreibondarev

**Comment by alchaplinsky** [comment]
@mathieuripert This is a good workaround for handling this. However, this response reconstruction from chunks should be done on a lower level. Ideally `ruby-openai` should return a response regardless if a block was passed or not. But if we're fixing it in langchainrb the LLM class should handle this somewhere here: https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/llm/openai.rb#L139
`Conversation`  is a high-level concept and it should not deal with blocks and chunks and reconstructing a response from the API.

**Comment by mathieuripert** [comment]
> @mathieuripert This is a good workaround for handling this. However, this response reconstruction from chunks should be done on a lower level. Ideally `ruby-openai` should return a response regardless if a block was passed or not. But if we're fixing it in langchainrb the LLM class should handle this somewhere here: https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/llm/openai.rb#L139 `Conversation` is a high-level concept and it should not deal with blocks and chunks and reconstructing a response from the API.

Hi @alchaplinsky ! It makes total sense. Indeed, `ruby-openai` does not return any response when a stream block is passed. And I agree, we should handle it `Langchain::LLM::OpenAI`. I can take a look at it except if you want to do it.

File: lib/langchain/llm/openai.rb [hunk 1]
```diff
@@ -238,5 +240,22 @@ def extract_response(response)
       results = response.dig("choices").map { |choice| choice.dig("message", "content") }
       (results.size == 1) ? results.first : results
     end
+
+    def response_from_chunks
+      {
+        "id" => @response_chunks.first&.dig("id"),
+        "object" => @response_chunks.first&.dig("object"),
+        "created" => @response_chunks.first&.dig("created"),
+        "model" => @response_chunks.first&.dig("model"),
```
andreibondarev, line 249: Could probably simplify this a tad with:
```
@response_chunks
 .first&
 .slice("id", "object", "created", "model")
 .merge({
   "choices" => [
   ...
```

andreibondarev (approved):

andreibondarev:
@mathieuripert I like this, looks good. @alchaplinsky Want to take another glance?

alchaplinsky:
@mathieuripert  Alright, this looks good and works for a single choice in the API response. However, if you pass an `n` parameter that is greater that 1 you would get more choices coming in chunks. 

So if you want to make this solution more robust and handle this case, I'd recommend to change `response_chunks` to a Hash instead of Array, and record chunks based on the `index` from a response. And then assemble all the choices from arrays of chunks for each index.

This, however, might be done in a follow up PR if feel like submitting another one.

mathieuripert:
> @mathieuripert Alright, this looks good and works for a single choice in the API response. However, if you pass an `n` parameter that is greater that 1 you would get more choices coming in chunks.
> 
> So if you want to make this solution more robust and handle this case, I'd recommend to change `response_chunks` to a Hash instead of Array, and record chunks based on the `index` from a response. And then assemble all the choices from arrays of chunks for each index.
> 
> This, however, might be done in a follow up PR if feel like submitting another one.

Agree. Ideally wee need to support for `n>1` as well. But there is no regression with proposed change as the current streaming behavior only supports single choice.
It will be a good follow up PR though

mathieuripert:
> > @mathieuripert Alright, this looks good and works for a single choice in the API response. However, if you pass an `n` parameter that is greater that 1 you would get more choices coming in chunks.
> > So if you want to make this solution more robust and handle this case, I'd recommend to change `response_chunks` to a Hash instead of Array, and record chunks based on the `index` from a response. And then assemble all the choices from arrays of chunks for each index.
> > This, however, might be done in a follow up PR if feel like submitting another one.
> 
> Agree. Ideally wee need to support for `n>1` as well. But there is no regression with proposed change as the current streaming behavior only supports single choice. It will be a good follow up PR though

Actually, my bad. I just realized that with `n>1` we get all choices coming in chunks...

lib/langchain/llm/openai.rb
```diff
@@ -181,8 +180,11 @@ def compose_parameters(model, params, &block)
       parameters = default_params.merge(params)
 
       if block
+        @response_chunks = []
         parameters[:stream] = proc do |chunk, _bytesize|
-          yield chunk.dig("choices", 0)
+          chunk_content = chunk.dig("choices", 0)
+          @response_chunks << chunk
```
andreibondarev, line 186: It looks like in order to accomodate for `n > 1`, we'd need reconstruct the chunks based on their index:
```ruby
irb(main):007:1* response = llm.chat(prompt: "hello! How are you?", n: 2, temperature: 1)  do |chunk|
irb(main):008:1*   puts chunk
irb(main):009:0> end
{"index"=>0, "delta"=>{"role"=>"assistant", "content"=>""}, "finish_reason"=>nil}
{"index"=>0, "delta"=>{"content"=>"Hello"}, "finish_reason"=>nil}
{"index"=>1, "delta"=>{"role"=>"assistant", "content"=>""}, "finish_reason"=>nil}
{"index"=>1, "delta"=>{"content"=>"Hello"}, "finish_reason"=>nil}
{"index"=>0, "delta"=>{"content"=>"!"}, "finish_reason"=>nil}
{"index"=>1, "delta"=>{"content"=>"!"}, "finish_reason"=>nil}
{"index"=>0, "delta"=>{"content"=>" I"}, "finish_reason"=>nil}
...
```

mathieuripert:
> > > > @mathieuripert Alright, this looks good and works for a single choice in the API response. However, if you pass an `n` parameter that is greater that 1 you would get more choices coming in chunks.
> > > > So if you want to make this solution more robust and handle this case, I'd recommend to change `response_chunks` to a Hash instead of Array, and record chunks based on the `index` from a response. And then assemble all the choices from arrays of chunks for each index.
> > > > This, however, might be done in a follow up PR if feel like submitting another one.
> > > 
> > > 
> > > Agree. Ideally wee need to support for `n>1` as well. But there is no regression with proposed change as the current streaming behavior only supports single choice. It will be a good follow up PR though
> > 
> > 
> > Actually, my bad. I just realized that with `n>1` we get all choices coming in chunks...
> 
> We can merge this PR now in the meantime if that's what you'd prefer.

Yes let's merge it if of for you guys ! I'll follow up with another PR to handle the `n >1` case. Thanks!

Issue #365, opened 2023-10-22
OpenAI#complete does not appear to allow max_tokens to be set


bricolage:
It looks like `OpenAI#complete` forces `max_tokens` to be the max available tokens based on the prompt / context.

I have to read more of the code to see how parameter defaults are being set, but it seems like this one might be as simple as changing the assignment to the "or-equals" operator, allowing it to be set only if it wasn't passed in as a param (that would be expected behavior, to my mind).

e.g.
```
parameters[:max_tokens] ||= validate_max_tokens(parameters[:messages], parameters[:model])
```

andreibondarev:
@bricolage Would you be open to submitting a PR?

bricolage:
@andreibondarev yeah for sure, should be able to do in the next few days

andreibondarev:
@bricolage We probably need to fix this for all other LLMs as well?

bricolage:
@andreibondarev yeah, I'm not super familiar with the other LLM APIs but am happy to tackle in another PR if we don't have any other devs who are pseudo-maintainers of the LLM wrappers.

bricolage:
@andreibondarev just update, I'll get support in the other LLM validators for their version of `:max_tokens` over the next few days.

andreibondarev:
> @andreibondarev just an update, I'll get support in the other LLM validators for their version of `:max_tokens` over the next few days.

That sounds great, thank you!

PR #388, opened 2023-11-12
Support a :max_tokens param in OpenAIValidator.validate_max_tokens

bricolage:
# Motivation

Two of OpenAI's API endpoints (chat and completions) take a `:max_tokens` param, but these are ignored in the current token validation / calculation method `BaseValidator.validate_max_tokens` (that method name is slightly misleading as it does more than validate it, but that's for another PR).

# Changes

Fortunately the changes were minimal. `BaseValidator.validate_max_tokens` already takes an options param, and I just worked on this method in #379. Like in #379 we want to take the lower of two token limits. In this case it's between the max calculated from the leftover tokens and `options[:max_tokens]`. I added two specs for when it's passed in (one for when it's lower, one for when it's higher than the calculated max).

Addresses #365

### DIFF
--- a/lib/langchain/llm/openai.rb
+++ b/lib/langchain/llm/openai.rb
@@ -69,7 +69,7 @@ module Langchain::LLM
       return legacy_complete(prompt, parameters) if is_legacy_model?(parameters[:model])
 
       parameters[:messages] = compose_chat_messages(prompt: prompt)
-      parameters[:max_tokens] = validate_max_tokens(parameters[:messages], parameters[:model])
+      parameters[:max_tokens] = validate_max_tokens(parameters[:messages], parameters[:model], parameters[:max_tokens])
 
       response = with_api_error_handling do
         client.chat(parameters: parameters)
@@ -131,7 +131,7 @@ module Langchain::LLM
       if functions
         parameters[:functions] = functions
       else
-        parameters[:max_tokens] = validate_max_tokens(parameters[:messages], parameters[:model])
+        parameters[:max_tokens] = validate_max_tokens(parameters[:messages], parameters[:model], parameters[:max_tokens])
       end
 
       response = with_api_error_handling { client.chat(parameters: parameters) }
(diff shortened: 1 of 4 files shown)


lib/langchain/llm/openai.rb
```diff
@@ -69,7 +69,7 @@ def complete(prompt:, **params)
       return legacy_complete(prompt, parameters) if is_legacy_model?(parameters[:model])
 
       parameters[:messages] = compose_chat_messages(prompt: prompt)
-      parameters[:max_tokens] = validate_max_tokens(parameters[:messages], parameters[:model])
+      parameters[:max_tokens] = validate_max_tokens(parameters[:messages], parameters[:model], parameters[:max_tokens])
```
andreibondarev, line 72:
```ruby
max_tokens = [max_tokens, options[:max_tokens]].min
```
Do you think this ^^ logic should actually live inside of this method and not inside the base_validator or the openai_validator?

For example we could have something like this:
```ruby
parameters[:max_tokens] = [
  parameters[:max_tokens],
  validate_max_tokens(parameters[:messages], parameters[:model])
].min
```

I guess what's missing with this approach is raising an error?

bricolage: @andreibondarev if `:max_tokens` is unique to OpenAI then definitely no need to bleed into the `BaseValidator`. Otherwise it will mean each LLM endpoint will need to implement similar code. What do you think?

bricolage: @andreibondarev also what error condition did you have in mind? e.g. if `:max_tokens` is passed in as a non-number value? Or possibly as a negative value?

bricolage: @andreibondarev check out the latest commit. I moved it into the `OpenAIValidator` subclass.

andreibondarev: > @andreibondarev also what error condition did you have in mind? e.g. if `:max_tokens` is passed in as a non-number value? Or possibly as a negative value?

This is the error that I was referring to: https://github.com/andreibondarev/langchainrb/pull/388/files#diff-79c9e673e647cc360e11c2946b3a4def6a5dfe52480860dc9ccd2101906902a9R30

andreibondarev (approved):

andreibondarev:
@bricolage Thank you!

PR #389, opened 2023-11-13
Updated Qdrant add_texts method to allow dynamic payloads

roshan92:
The key changes:
- Accept a payloads argument to allow custom payloads
- Use the passed payload if provided, else default to {content: text}

### DIFF
--- a/lib/langchain/vectorsearch/qdrant.rb
+++ b/lib/langchain/vectorsearch/qdrant.rb
@@ -49,9 +49,9 @@ module Langchain::Vectorsearch
 
       Array(texts).each_with_index do |text, i|
         id = ids[i] || SecureRandom.uuid
-        batch[:ids].push(id)        
+        batch[:ids].push(id)
         batch[:vectors].push(llm.embed(text: text).embedding)
-        batch[:payloads].push(payload.present? ? {content: text}.merge!(payload) : {content: text})
+        batch[:payloads].push({content: text}.merge(payload))
       end
 
       client.points.upsert(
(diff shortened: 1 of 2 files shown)


roshan92 (commented):
Updated to use a ternary operator to handle payloads in one line. Then, merges custom payload if present, else just uses default.

roshan92 (commented):
Updated payload to accept hash.

lib/langchain/vectorsearch/qdrant.rb
```diff
@@ -44,14 +44,14 @@ def find(ids: [])
...
+    def add_texts(texts:, ids: [], payload: {})
       batch = {ids: [], vectors: [], payloads: []}
 
       Array(texts).each_with_index do |text, i|
         id = ids[i] || SecureRandom.uuid
-        batch[:ids].push(id)
+        batch[:ids].push(id)        
         batch[:vectors].push(llm.embed(text: text).embedding)
-        batch[:payloads].push({content: text})
+        batch[:payloads].push(payload.present? ? {content: text}.merge!(payload) : {content: text})
```
andreibondarev, line 54: I think this can be simplified with:
```ruby
{ content: text }.merge(payload)
```
Because even if `payload` is `{}`, it'll just work.

andreibondarev:
@roshan92 Could you please add a test and fix the linter errors? Thank you!

lib/langchain/vectorsearch/qdrant.rb
roshan92: Yes, you're right on that. I will fix that.

roshan92 (commented):
added spec for merging provided payload with default payload

roshan92 (commented):
Fixed linter error and simplified logic

PR #398, opened 2023-11-23
 Adding support for Embeddings for Vertex AI - one step at a time

palladius:
Hi Andre - I need to run for dinner, but happy to iterate and get to where you want this to be!
Or we can just drop it and you rewrite it better :) you tell me.

## REPRO 

```
# File ZZZ-riccardo.rb

######################################################################

require 'google/cloud/ai_platform'
require 'google/cloud/ai_platform/v1'
require "google/apis/aiplatform_v1"

llm = Langchain::LLM::GoogleVertexAi.new(
  #api_key_file: ENV['GOOGLE_CLOUD_VERTEX_AI_KEY'], # useless
  project_id: 'ricc-genai'
)
llm.embed(text: "foo bar")
ret = llm.embed(text: "foo bar")
#puts ret.embeddings rescue "No Embed mthod"
puts ret.class

exit 0

```

Then I issue this command: 

```
(base) ricc-macbookpro3:langchainrb-palladius ricc$ cat ZZZ-riccardo.rb | bin/console



# =>

# #<Langchain::LLM::GooglePalmResponse:0x0000000127eb9e48
# @model="textembedding-gecko",
# @prompt_tokens=nil,
# @raw_response=
#  #<Google::Apis::AiplatformV1::GoogleCloudAiplatformV1PredictResponse:0x0000000127ecad88
#   @metadata={"billableCharacterCount"=>6},
#   @predictions=
#    [{"embeddings"=>
#       {"statistics"=>{"token_count"=>3, "truncated"=>false},
#        "values"=>
#         [9.821750523997252e-08,
#          0.008036692626774311,
#          0.025743281468749046,
#       [...] 768 of these :)
#          0.011326898820698261,
#          -0.044290922582149506,
#          0.0052716382779181]}}]>>

```


palladius:
Note the code is pretty bad for the moment - I just wanted to show how I would approach it.
I fixed locally the hardcoded  project_id if anyone wants to see it.
My idea was to push a working version with JUST embeddings and then start building everything else one step at a time. 
Some questions:

* does it make sense to branch Vertex AI vs Palm API or should we use a single (bloated) version?
* is it ok to implement one functionality at a time, or a provider should be all or nothing?
* if we have TWO, can I keep `GooglePalmResponse` fior VertexAI or should I spawn a copy into some `VertexAiResponse`? one is more elegant, one is more practical and immediately DRY. 

Note (part of) VertexAI Embeddings implementation in python : https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/embeddings/vertexai.py

andreibondarev:
@palladius I'm going to close your PR but merge my cleaned up PR with your changes: https://github.com/andreibondarev/langchainrb/pull/402/

PR #401, opened 2023-11-29
Llama cpp version update

spikex:
The current version of the [llama_cpp](https://github.com/yoshoku/llama_cpp.rb) gem is 0.9.4. langchainrb was using 0.3.7. The llama_cpp API has changed (because the underlying [llama.cpp](https://github.com/ggerganov/llama.cpp) API is in flux), so it's not a simple upgrade.

I've kept the `Langchain::LLM::LlamaCpp` API the same, however this **is a breaking change** for this LLM as the model file format that llama.cpp uses has changed. The model files now need be in GGUF format.

The specs pass and it can also be tested by running `ruby -Ilib -rllama_cpp examples/llama_cpp.rb`


spikex:
@andreibondarev All set, let me know if it looks good.

andreibondarev:
@spikex Thank you!

andreibondarev:
@spikex Are you actually using the Langchain::LLM::LlamaCpp class in your applications? I'm debating getting rid of it in favor of the Ollama integration.

spikex:
@andreibondarev I've used it for some experiments, but am not using it in production. I won't miss it.

PR #405, opened 2023-12-01
add epsilla db connector

tonyyanga:
This PR adds a database connector to [Epsilla](https://github.com/epsilla-cloud/vectordb), an open source vector database. It allows langchainrb users to use it with the unified interface that currently supports a few others.

### Testing
I added unit tests that run along with the official Epsilla docker container.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,5 +15,3 @@ gem "rubocop"
 
 # Temporary fix until https://github.com/github/graphql-client/pull/314 is merged
 gem "graphql-client", git: "https://github.com/rmosolgo/graphql-client.git", branch: "start-migrating"
-
-gem "epsilla-ruby", "~> 0.0.4"
--- a/Gemfile
+++ b/Gemfile
@@ -15,3 +15,5 @@ gem "rubocop"
 
 # Temporary fix until https://github.com/github/graphql-client/pull/314 is merged
 gem "graphql-client", git: "https://github.com/rmosolgo/graphql-client.git", branch: "start-migrating"
+
+gem "epsilla-ruby", "~> 0.0.4"
--- a/lib/langchain/vectorsearch/epsilla.rb
+++ b/lib/langchain/vectorsearch/epsilla.rb
@@ -3,6 +3,7 @@
 require "securerandom"
 require "json"
 require "timeout"
+require "uri"
 
 module Langchain::Vectorsearch
   class Epsilla < Base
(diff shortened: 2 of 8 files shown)


Gemfile
```diff
@@ -15,3 +15,5 @@ gem "rubocop"
 
 # Temporary fix until https://github.com/github/graphql-client/pull/314 is merged
 gem "graphql-client", git: "https://github.com/rmosolgo/graphql-client.git", branch: "start-migrating"
+
+gem "epsilla-ruby", "~> 0.0.4"
```
andreibondarev, line 19: If it's just in the `langchain.gemspec` file -- it's sufficient.

lib/langchain/vectorsearch/epsilla.rb
```diff
@@ -0,0 +1,131 @@
...
+    #
+    # Initialize Epsilla client
+    # @param protocol [String] The protocol to use, e.g. http or https
+    # @param host [String] The host to connect to
+    # @param db_name [String] The name of the database to use
+    # @param db_path [String] The path to the database to use
+    # @param index_name [String] The name of the Epsilla table to use
+    # @param llm [Object] The LLM client to use
+    # @param port [Integer] The port to connect to, default 8888
+    def initialize(protocol:, host:, db_name:, db_path:, index_name:, llm:, port: 8888)
```
andreibondarev, line 26: I would kind of prefer that the initializer just accepts `url:, llm:` params.

Would it then be possible to break the url into protocol, host, port, db_name, path, etc as needed?

lib/langchain/vectorsearch/epsilla.rb
```diff
@@ -0,0 +1,131 @@
...
+    def destroy_default_schema
+      status_code, response = @client.database.drop_table(@table_name)
+      raise "Failed to drop table: #{response}" if status_code != 200
+
+      response
+    end
+
+    # Add a list of texts to the database
+    # @param texts [Array<String>] The list of texts to add
+    def add_texts(texts:)
```
andreibondarev, line 76: Is it possible to pass our own `ids:`?

tonyyanga: Good call. Added support for passing in custom IDs.

lib/langchain/vectorsearch/epsilla.rb
tonyyanga: It's quite natural to use `url` to encode `protocol`, `host` and `port`, so I have made the change to simplify the params list.

However, Epsilla does not have a canonical way to represent `db_name` and `db_path` in a URL. Compared to other databases, it is a bit clunky. I thought it would be better to keep it explicit and avoid confusions.

What do you think?

tonyyanga:
Thanks for the review @andreibondarev ! Made changes based on your feedback. Please let me know if you have any other concerns.

andreibondarev:
@tonyyanga Solid! Thank you! 👍🏻

PR #406, opened 2023-12-02
 Adding Complete and Summarize for Vertex AI, along with a number of …

palladius:
…enhancements

* Adding `summarize()` and `complete()` to Vertex AI
* Adding opinionated values from public documentation for Temperature/TopP/TopK/MaxOutputToken
* Adding convenience methods to `GoogleVertexAIResponse`. I took the liberty to add a .summary() which seems more idiomatic. I wonder how you implemented the dig() which works for you on Plam API but not for me in Vertex.
* renaming `max_tokens` to `max_output_tokens` . It definitely works on the Google side, please watch its ok with the rest.
* Adding `GOOGLE_CLOUD_CREDENTIALS` to .env/example since its automatically assimilated and used by ruby GCP auth module  - if set.

As usual, the code is a bit bloated, so feel free to remove all my debug puts() :)


palladius:
Andrei, anything I can do?

PR #411, opened 2023-12-06
Fix README examples

mattlindsey:
Add .completion to response examples, and default_options example with google.


andreibondarev:
@mattlindsey Thank you!

PR #419, opened 2023-12-14
Fix OpenAIValidator when messages: [...] is passed in

andreibondarev:


lib/langchain/utils/token_length/openai_validator.rb
```diff
@@ -75,6 +75,47 @@ def self.validate_max_tokens!(content, model_name, options = {})
...
+          encoding = Tiktoken.encoding_for_model(model_name)
+
+          if ["gpt-3.5-turbo-0613", "gpt-3.5-turbo-16k-0613", "gpt-4-0314", "gpt-4-32k-0314", "gpt-4-0613", "gpt-4-32k-0613"].include?(model_name)
+            tokens_per_message = 3
+            tokens_per_name = 1
+          elsif model_name == "gpt-3.5-turbo-0301"
+            tokens_per_message = 4  # every message follows {role/name}\n{content}\n
+            tokens_per_name = -1  # if there's a name, the role is omitted
+          elsif model_name.include?("gpt-3.5-turbo")
+            puts "Warning: gpt-3.5-turbo may update over time. Returning num tokens assuming gpt-3.5-turbo-0613."
```
mazenkhalil, line 96: Just wondering why puts is being used here? Should this be logged instead?

lib/langchain/utils/token_length/openai_validator.rb
```diff
@@ -75,6 +75,47 @@ def self.validate_max_tokens!(content, model_name, options = {})
           max_tokens = super(content, model_name, options)
           [options[:max_tokens], max_tokens].reject(&:nil?).min
         end
+
+        # Copied from https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb
```
mazenkhalil, line 79: This is just an example from OpenAI. What about the newer models? e.g gpt-3.5-turbo-1106

mazenkhalil:
I've added some comments to the changes done here. I am not sure if such changes are suitable for production!

lib/langchain/utils/token_length/openai_validator.rb
mazenkhalil: Also the calculation doesn't consider the tools/functions if being used.

andreibondarev: > Also the calculation doesn't consider the tools/functions if being used.

I would love to have this fixed!

lib/langchain/utils/token_length/openai_validator.rb
andreibondarev: I think it's okay as just a comment. I'm not sure it needs to be logged as a warning or anything.

PR #193, opened 2023-06-18
Handle exceptions in Weather Tool

mattlindsey:
Fixes #175 and handles units better.

There are some quirks in the open source weather gem that require this workaround.  I inquired with the gem owner about fixing them, but it would require substantial work to fix or to update the gem to the latest Open Weather API 3.0.

PR #422, opened 2023-12-19
Assistants

andreibondarev:


mattlindsey:
Hi. I tried this PR and looks pretty good, but weather tool still blows up.  #193 looks a little messy but fixes it.

mattlindsey:
Also, the README example works, but the tool with NAME of `google_search` is called even though it says call `<tool>weather</tool>`.  Shouldn't the one with a matching name get called first if it exists?  Or is the llm deciding?

lib/langchain/assistants/prompts/tools_prompt.yaml
```diff
@@ -0,0 +1,16 @@
...
+  You have access to the following tools:
+
+  {tools}
+
+  In order to use a tool, you can use <tool></tool> and <tool_input></tool_input> tags. The tool output will be provided to you in a follow-up chat message.
+
+  For example, if you have a tool called "google_search" that could run a Google search, in order to search for the weather in SF you would respond: <tool>google_search</tool><tool_input>weather in SF</tool_input>
+
+  user: What is the current weather in Boston, MA?
+  assistant: <tool>google_search</tool><tool_input>Boston, MA</tool_input>
```
mattlindsey, line 14: Minor point, but I think this example should be `<tool_input>weather in Boston, MA</tool_input>`, since the google_search tool wouldn't know what you're querying about. (It's correct in line 11.)
Another minor point is, if you add the `weather` tool, the llm might get confused about which tool to call. Maybe use an example that should always go to google_search.

mattlindsey: Openai recommends priming an assistant with faked examples, so I think <example_tool> might work well.
Also a few examples with  "role": "system", "name":"example_user".

mattlindsey (approved):
Looks good to me. Maybe add 'legacy' or 'deprecated' message at top of agents.md, and replace words 'Chat Bots' with 'Assistants' in README.md.

PR #430, opened 2023-12-27
Update ci.yml

andreibondarev:


andreibondarev:
@tonyyanga I had to lock the version to 0.3.1 because the newest build was breaking the tests.

PR #431, opened 2023-12-27
Add Ruby 3.3 support

andreibondarev:


rickychilcott:
Ruby 3.3 tests run locally for me (off of `main`)

andreibondarev:
@rickychilcott Any chance you can run bundle install on 3.3 and commit `Gemfile.lock` to this branch? For some reason I've been having some issues with the tiktoken gem.

rickychilcott:
Ok. You're right. I've got the same problem. What I was really saying in my previous comment is that Ruby 3.3 seemed to be working for me -- I guess I'm not using OpenAI's API, so I wasn't running into the tiktoken issue.

It looks like https://github.com/IAPark/tiktoken_ruby/pull/21 and https://github.com/IAPark/tiktoken_ruby/issues/20 will need resolved before 3.3 support can be completed.

prdanelli:
In a recent comment regarding Ruby 3.3 support, it seems that [the tiktoken_ruby](https://github.com/IAPark/tiktoken_ruby/pull/21#issuecomment-1938163912) owner is burnt out from maintaining the Gem and wont be able to prioritise support. 

They have just merged in a patch for Ruby 3.3 support, but if that might well be the last update. 

[Thoughtbot](https://github.com/thoughtbot/tiktoken_ruby) have a fork, but who knows if they're able to support it long term.

andreibondarev:
Looks like there's an issue with the `safe_ruby` gem if anyone wants to take a look at it...

mattlindsey:
> Looks like there's an issue with the `safe_ruby` gem if anyone wants to take a look at it...

Yeah. Commands like `SafeRuby.eval('[1,2,3].map{ |n| n + 1 }')` no longer work in the console and give an `incompatible marshal file format` error.  I think it might be the dependencies for the safe_ruby gem need to be updated.  Last update was 7 years ago.  I'll see if I can do that...

mattlindsey:
> @andreibondarev This PR fixes safe_ruby, but may not be correct, so I'm hoping somebody reviews it: [ukutaht/safe_ruby#5](https://github.com/ukutaht/safe_ruby/pull/5).

@andreibondarev An option to get ruby 3.3 supported would be for someone to fork safe_ruby or accept 'transfer' of it (as suggested by the owner of the repo), apply the PR changes above, then include the new gem/repo in langchainrb.  The safe_ruby gem still has problems (which could be raised as Issues in the new repo), but I'm just giving an alternative to removing it entirely as I proposed in a different PR.

andreibondarev:
@mattlindsey I figured a better approach here... excluding the RubyCodeIntepreter tool for Ruby 3.3 and above since the `safe_ruby` gem throws an error then.

PR #433, opened 2023-12-28
Return any database error from the tool

### LINKED ISSUE #432: Handling of tool errors (e.g. Database)
Example below shows where a Tool raises an error, but the Assistant doesn't recognize it and happily gives wrong answer.   Maybe the Assistant can at least acknowledge that an error happened, and in the specific case below if the Assistant gets access to a 'schema' function/tool, it should work better.

image

mattlindsey:
Fixes #432


mattlindsey:
Hi @andreibondarev.  Can you merge this, or give review comments?

mattlindsey:
Note that another solution is to take out the `rescue` and catch the error in assistant.rb:

```ruby
       begin
          output = tool_instance.execute(**tool_arguments)
        rescue Exception => e
          output = e.message
        end
 ```
 
 https://github.com/andreibondarev/langchainrb/blob/64c452c09c4e1935bd3b92ef5328ddb4ae59e7f6/lib/langchain/assistants/assistant.rb#L149

mattlindsey:
@andreibondarev  I updated this PR to catch exceptions in assistant.rb, which seems to work well.  I tested with database and weather tools and it handled exceptions much better.

mattlindsey:
I think these changes should still be made, but I should probably re-do this PR since it looks like database.rb was moved to a subdirectory.
And rename the PR to 'Submit tool exceptions to Assistant for reconsideration by LLM.'

PR #439, opened 2024-01-09
Handle openai stop param

### LINKED ISSUE #438: Agents error with stop_sequences
I think the Agents under lib/langchain/agent are broken.  I tried the example in README and got error referencing `stop_sequences`:

image

mattlindsey:
Fixes #438

(Maybe a spec should have failed before this fix, or an integration test as proposed here #145 )


andreibondarev:
@mattlindsey Nice fix, thank you!

PR #442, opened 2024-01-15
Fix OpenAI Invalid value for "tool_choice"

mengqing:
This is to fix the following error, only pass tool_choice if tools are present

```
Langchain::LLM::ApiError: OpenAI API error: Invalid value for 'tool_choice': 'tool_choice' is only allowed when 'tools' are specified.
```

### DIFF
--- a/lib/langchain/llm/openai.rb
+++ b/lib/langchain/llm/openai.rb
@@ -133,7 +133,7 @@ module Langchain::LLM
       parameters[:temperature] = temperature if temperature
       parameters[:top_p] = top_p if top_p
       parameters[:tools] = tools if tools.any?
-      parameters[:tool_choice] = tool_choice if tool_choice
+      parameters[:tool_choice] = tool_choice if tools.any? && tool_choice
       parameters[:user] = user if user
 
       # TODO: Clean this part up


andreibondarev:
@mengqing How did you encounter this error?

mengqing:
If I run the `Assistants` example without specifying tools, eg

```
assistant = Langchain::Assistant.new(
  llm: llm,
  thread: thread,
  instructions: "You are a Meteorologist Assistant that is able to pull the weather for any location"
)
assistant.add_message_and_run content: "What's the weather in New York City?"
```
Then it returns the above error.

The culprit is here https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/assistants/assistant.rb#L131

andreibondarev:
@mengqing Do you think your fix is better than on [that line](https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/assistants/assistant.rb#L131) doing:
```ruby
tool_choice: tools.any? ? "auto" : "none"
```

mengqing:
The error still occurs even if setting `tool_choice` to `none` when `tools` is blank.

eg:

```
curl https://api.openai.com/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -d '{
    "model": "gpt-3.5-turbo",
    "messages": [
      {
        "role": "system",
        "content": "You are a helpful assistant."
      },
      {
        "role": "user",
        "content": "Hello!"
      }
    ],
    "tool_choice": "none"
  }'
```

andreibondarev:
@mengqing Then I'm thinking the proper thing to do would be to do this:
```ruby
      params = {messages: thread.openai_messages}
      # TODO: Not sure that tool_choice should always be "auto"; Maybe we can let the user toggle it.
      params.merge!(tools: tools.map(&:to_openai_tool), tool_choice: "auto") if tools.any?

      llm.chat(**params)
```

What do you think?

mengqing:
> @mengqing Then I'm thinking the proper thing to do would be to do this:
> 
> ```ruby
>       params = {messages: thread.openai_messages}
>       # TODO: Not sure that tool_choice should always be "auto"; Maybe we can let the user toggle it.
>       params.merge!(tools: tools.map(&:to_openai_tool), tool_choice: "auto") if tools.any?
> 
>       llm.chat(**params)
> ```
> 
> What do you think?

I think because the issue comes from the underlying OpenAI's API, someone is still able to call the `llm.chat` with the following args and be faced with the same issue

```
llm.chat({..., tool_choice: "auto"})
```

However this isn't a big issue because the issue here can be avoided by validating the arguments passed to the method, where as it is not possible currently with the Assistant API.

So I think this method is also good

andreibondarev:
@mengqing I ended up fixing it slightly differently. I'll publish a new gem version in a few. Thank you for finding the bug!

PR #453, opened 2024-01-29
Adding support for new embedding models + adding missing base_uri parameter for pinecone

iotventurefelix:
Adding support for text-embedding-3-large and text-embedding-3-small. 

PR contains a workaround for token length calculation, since TikToken does not yet support these new embedding models. 

Also fixed a missing "base_uri" parameter for Pinecone, which might be neccessary in some setups.


andreibondarev:
@felix-letkemann The `base_uri` is optional, right?

felix-letkemann:
@andreibondarev Exactly! 

I was using this version of the pinecone lib by the way: 
gem 'pinecone', git: 'https://github.com/w-zygmuntowicz/pinecone.git', branch: 'feature/serverless'

Might be usefull in the same setup

PR #460, opened 2024-02-01
Require Ollama LLM provider dependency

### LINKED ISSUE #458: langchain/llm/ollama.rb NameError: uninitialized constant Langchain::LLM::Ollama::Faraday
- In the Ollama client, Faraday is assumed to be autoloaded, which is not likely the case when outside of Rails
- Faraday also does not show up in the gem dependencies, and therefor must be installed separately, even though the constant is directly referenced in [this code](https://github.com/andreibondarev/langchainrb/blob/005c39b38b8d15bd0d0ed40cb819fb61d749d24d/lib/langchain/llm/ollama.rb#L217).

I see that Faraday appears in the Gemfile for the project, so tests pass when doing development, but do not work through gem distribution. Additionally, `require 'faraday'` is in the test file, but should also be in the client file to play nicely without Rails.

```
▶ gem install pry langchainrb
Fetching pry-0.14.2.gem
Successfully installed pry-0.14.2
Parsing documentation for pry-0.14.2
Installing ri documentation for pry-0.14.2
Done installing documentation for pry after 0 seconds
Fetching langchainrb-0.9.0.gem
Successfully installed langchainrb-0.9.0
Parsing documentation for langchainrb-0.9.0
Installing ri documentation for langchainrb-0.9.0
Done installing documentation for langchainrb after 0 seconds
2 gems installed

~
▶ pry
[1] pry(main)> require 'langchainrb'
=> true
[2] pry(main)> llm = Langchain::LLM::Ollama.new(url: 'http://localhost:11434')
=> #<Langchain::LLM::Ollama:0x000000011fb59918 @url="http://localhost:11434">
[3] pry(main)> llm.complete(prompt: "Hello world", model: "mistral")
NameError: uninitialized constant Langchain::LLM::Ollama::Faraday
from /Users/ttilberg/.asdf/installs/ruby/3.3.0/lib/ruby/gems/3.3.0/gems/langchainrb-0.9.0/lib/langchain/llm/ollama.rb:84:in `client'
[4] pry(main)> require 'faraday'
LoadError: cannot load such file -- faraday
from <internal:/Users/ttilberg/.asdf/installs/ruby/3.3.0/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:127:in `require'
```

ttilberg:
This change establishes Faraday as a required dependency for the Ollama LLM provider. Additionally, it starts to add documentation in the Readme that each of the LLMs may have their own dependencies that are not included by default.

Fixes #458

### DIFF
--- a/lib/langchain/llm/ollama.rb
+++ b/lib/langchain/llm/ollama.rb
@@ -22,6 +22,7 @@ module Langchain::LLM
     # @param default_options [Hash] The default options to use
     #
     def initialize(url:, default_options: {})
+      depends_on "faraday"
       @url = url
       @defaults = DEFAULTS.merge(default_options)
     end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -424,6 +424,7 @@ DEPENDENCIES
   elasticsearch (~> 8.2.0)
   epsilla-ruby (~> 0.0.4)
   eqn (~> 1.6.5)
+  faraday
   google-apis-aiplatform_v1 (~> 0.7)
   google_palm_api (~> 0.1.3)
   google_search_results (~> 2.0.0)
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ If bundler is not being used to manage dependencies, install the gem by executin
 
     gem install langchainrb
 
+Additional gems may be required when loading LLM Providers. These are not included by default so you can include only what you need.
+
 ## Usage
 
 ```ruby
--- a/langchain.gemspec
+++ b/langchain.gemspec
@@ -73,4 +73,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "sequel", "~> 5.68.0"
   spec.add_development_dependency "weaviate-ruby", "~> 0.8.10"
   spec.add_development_dependency "wikipedia-client", "~> 1.17.0"
+  spec.add_development_dependency "faraday"
 end


lib/langchain/llm/ollama.rb
```diff
@@ -22,6 +22,7 @@ class Ollama < Base
     # @param default_options [Hash] The default options to use
     #
     def initialize(url:, default_options: {})
+      depends_on 'faraday'
```
andreibondarev, line 25: Double quotes, otherwise the linter complains.

andreibondarev (approved):

andreibondarev:
Thank you for the fix. Just the linter error!

ttilberg:
Hi @andreibondarev,

After taking my initial stab at just `require`ing Faraday, I started to notice other infrastructure in place to help load required Gems using Langchain::DependencyHelper. I've mimicked how other LLM providers appear to be handling dependency resolution. However, I'm not sure I have this configured correctly:

If I comment out `require 'faraday'` in the spec, I'd expect the spec that references the Faraday constant to possibly fail if it hasn't been required yet, but I also get

```
  2) Langchain::LLM::Ollama#chat returns a chat completion
     Failure/Error:
       @client ||= Faraday.new(url: url) do |conn|
         conn.request :json
         conn.response :json
         conn.response :raise_error
       end

     NameError:
       uninitialized constant Langchain::LLM::Ollama::Faraday

             @client ||= Faraday.new(url: url) do |conn|
                         ^^^^^^^
     # ./lib/langchain/llm/ollama.rb:220:in `client'
     # ./spec/langchain/llm/ollama_spec.rb:43:in `block (3 levels) in <top (required)>'

```

I would have expected this to be resolved by the `depends_on` helper, do you agree?

lib/langchain/llm/ollama.rb
ttilberg: Hah, I caught that and re-force pushed right afterwards. You were too quick (I was too slow!) ;)

andreibondarev:
@ttilberg I suspect that you may need to add `faraday` to the gemspec file. 

Also -- does referencing the top-level namespace fix the issue?
```ruby
@client ||= ::Faraday.new
```

ttilberg:
@andreibondarev I've added Faraday to the Gemspec. I believe this brings parity to the Ollama provider now.

andreibondarev:
@ttilberg Let's ship it! 🚀 🚢

PR #469, opened 2024-02-14
Allow the use of OpenAI's rough token count

Sergio-e:
## Context
Noticed that the [tiktoken_ruby](https://github.com/IAPark/tiktoken_ruby) dependency isn't thread-safe. For us, this resulted in the Sidekiq process crashing consistently. [There is also an open issue regarding this problem](https://github.com/IAPark/tiktoken_ruby/issues/8)

## Proposal
Allow the use of [OpenAI's rough token count](https://github.com/alexrudall/ruby-openai?tab=readme-ov-file#counting-tokens) instead of using tiktoken ruby. The token count selection will be made when initializing the LLM:
```
    Langchain::LLM::OpenAI.new(
      api_key: openai_access_token,
      llm_options: {},
      default_options: { token_counter: :openai }
    )
```
tiktoken_ruby will be left as the default option.


andreibondarev:
@Sergio-e Do we still need this? I got rid of the automatic max_tokens validation via tiktoken. I think that feature was more annoying than helpful.

Sergio-e:
> @Sergio-e Do we still need this? I got rid of the automatic max_tokens validation via tiktoken. I think that feature was more annoying than helpful.

It seems not. The issue is now fixed on the latest Tiktoken ruby release

PR #478, opened 2024-02-22
Allow json-schema 4.1

AndyGauge:
In order to add to our gemset, we need langchainrb to loosen their dependency restrictions.  json-schema 4.0 was good but 4.1 is also compatible.  By reducing the zeros in the pessimistic lock, the minor version can rise without allowing 5.x versions.


andreibondarev:
@AndyGauge I think you have to bundle install and commit the lock file.

andreibondarev:
@AndyGauge I went ahead and did that for you here: https://github.com/andreibondarev/langchainrb/pull/480

PR #479, opened 2024-02-23
Fix for updating record on pinecone

ddyy:
When updating a record with pinecone db, I receive this error:

NoMethodError: undefined method `empty?' for nil:NilClass (NoMethodError)

      inputs["setMetadata"] = set_metadata unless set_metadata.empty?
                                                              ^^^^^^^
/ruby-3.2.2/gems/pinecone-0.1.71/lib/pinecone/vector.rb:51:in `update'
/ruby-3.2.2/gems/langchainrb-0.8.2/lib/langchain/vectorsearch/pinecone.rb:91:in `block in update_texts'

This PR fixes the missing metadata by setting it the same way in update_data as it's set in add_data


lib/langchain/vectorsearch/pinecone.rb
```diff
@@ -93,7 +93,7 @@ def update_texts(texts:, ids:, namespace: "", metadata: nil)
           namespace: namespace,
           id: ids[i].to_s,
           values: llm.embed(text: text).embedding,
-          set_metadata: metadata
+          set_metadata: metadata || {content: text}
```
andreibondarev, line 96: Can we optionally pass metadata if it's set? Otherwise not even pass this parameter. Would that work for your usecase?

ddyy: Yes, that works for me, it's not required for my use case, but I followed the format in add_texts in case it was needed. The metadata is only convenient to identify the vector entries when browsing from pinecone db directly, but is not functionally necessary.

andreibondarev:
@ddyy Specs are broken.

andreibondarev:
Closing due to inactivity.

PR #482, opened 2024-02-23
OutputFixingParser: call `chat` with `messages` instead of `prompt` parameter when original parser raises an exception

### LINKED ISSUE #481: OutputFixingParser uses invalid params for OpenAI chat completion (and most other LLMs)
I'm seeing this error trying to use the OutputFixingParser with OpenAI:
```
/usr/local/rvm/gems/default/gems/langchainrb-0.9.2/lib/langchain/llm/openai.rb:93:in `chat': unknown keyword: :prompt (ArgumentError)
/usr/local/rvm/gems/default/gems/langchainrb-0.9.2/lib/langchain/output_parsers/structured_output_parser.rb:78:in `rescue in parse': Failed to parse. Text: "```json (Langchain::OutputParsers::OutputParserException)
```

Using something like this:
```ruby
Langchain::OutputParsers::OutputFixingParser.from_llm(
  llm: Langchain::LLM::OpenAI.new(api_key: <api_key>),
  parser: Langchain::OutputParsers::StructuredOutputParser.from_json_schema(<some_schema>)
)
```

The OutputFixingParser is making this call, which assumes both that the parser will respond to `chat` and that `chat` accepts the `prompt` keyword, which means this will fail for most LLMs. 
```ruby
   def parse(completion)
      parser.parse(completion)
    rescue OutputParserException => e
      new_completion = llm.chat(
        prompt: prompt.format(
          instructions: parser.get_format_instructions,
          completion: completion,
          error: e
        )
      ).completion
      parser.parse(new_completion)
    end
```

It looks like only `Langchain::LLM::GooglePalm` supports `prompt` as a parameter to `chat`

viamin:
Most LLM classes don't support the prompt parameter in the chat method. (In fact, only GooglePalm does.) This will allow more LLMs to use an OutputFixingParser (including Azure, Ollama, OpenAI, and ... GooglePalm)

I believe this should be a backwards compatible change for anyone using this as is with GooglePalm, but it does change the behavior. 

Unclear why the existing specs didn't catch this. 

Fixes #481


andreibondarev:
@viamin Thank you for the PR! If you use OpenAI, I'd probably just use the JSON mode they've got.

PR #498, opened 2024-02-28
Fix allow rails 7.0.x versions to use gem

prdanelli:
I realise this one might be controversial, but limiting the Gem to the most recent Rails version isn't always possible. We are upgrading to Rails 7.1, but its not a quick process on a large monolith.


andreibondarev:
@prdanelli I'll merge your PR. We can't remove `activesupport` because there's some other gems that require it.

prdanelli:
Thank you @andreibondarev.

PR #502, opened 2024-02-28
Add Message class tests

mattlindsey:
Fixes class name at top of file.   
Skips final test for now.  I may need help with it.

### DIFF
--- a/spec/langchain/assistants/message_spec.rb
+++ b/spec/langchain/assistants/message_spec.rb
@@ -11,7 +11,7 @@ RSpec.describe Langchain::Message do
     context "when role and content are not nil" do
       let(:message) { described_class.new(role: "user", content: "Hello, world!", tool_calls: [], tool_call_id: nil) }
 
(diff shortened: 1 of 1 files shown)


mattlindsey:
@andreibondarev I fixed the final test, so I think this looks OK now.

spec/langchain/assistants/message_spec.rb
```diff
@@ -1,7 +1,41 @@
 # frozen_string_literal: true
 
-RSpec.describe Langchain::Thread do
-  it "raises an error if messages array contains non-Langchain::Message instance(s)" do
-    expect { described_class.new(messages: [Langchain::Message.new, "foo"]) }.to raise_error(ArgumentError)
+RSpec.describe Langchain::Message do
+  subject { described_class.new }
```
andreibondarev, line 4: Btw, in Rails, `subject` always defaults to `described_class.new` so you don't even have to declare it.

PR #504, opened 2024-02-29
Optionally pass a block to OpenAI::Client.new

drnic:
This block is ultimately passed to Faraday.

For example, to enable low-level faraday logging:

```ruby
llm = Langchain::LLM::OpenAI.new(api_key: ENV[OPENAI_API_KEY]) do |f|
  f.response :logger, Logger.new(), bodies: true
end
```

See more at https://github.com/alexrudall/ruby-openai?tab=readme-ov-file#verbose-logging

### DIFF
--- a/lib/langchain/llm/openai.rb
+++ b/lib/langchain/llm/openai.rb
@@ -34,10 +34,10 @@ module Langchain::LLM
     #
     # @param api_key [String] The API key to use
     # @param client_options [Hash] Options to pass to the OpenAI::Client constructor
-    def initialize(api_key:, llm_options: {}, default_options: {})
+    def initialize(api_key:, llm_options: {}, default_options: {}, &llm_block)
       depends_on "ruby-openai", req: "openai"
 
-      @client = ::OpenAI::Client.new(access_token: api_key, **llm_options)
+      @client = ::OpenAI::Client.new(access_token: api_key, **llm_options, &llm_block)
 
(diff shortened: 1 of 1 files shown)


andreibondarev:
@drnic Is the logger the only thing that you think you'd need to pass to the Faraday client? We also have this global logging option: https://github.com/andreibondarev/langchainrb?tab=readme-ov-file#logging. I wonder if it's better to have this global flag toggle logging in Faraday here and elsewhere. What's your thoughts?

sergiobayona:
I ran into the same issue as @drnic where I could not pass a block to the openai client. Then I saw you mentioned the global logging flag, which I wasn't aware of before. However I still think it would be a nice addition to be able to pass the block.

sergiobayona:
yeah so used `Langchain.logger.level = :debug` and ran some api calls. I was expecting to see the request/response calls logged but that was not the case. A stronger case for merging this.

codenamev:
We have a use-case also requiring some customizations to the client.  Could we instead consider exposing a writer to `:client` itself (and maybe allowing setting in initialization)?  I think that would also solve the issue mentioned here as well.

andreibondarev:
OpenAI errors are funneled to the Langchain.logger now: https://github.com/patterns-ai-core/langchainrb/blob/0.19.0/lib/langchain/llm/openai.rb#L39.

Thank you for the PR though. Closing for now.

PR #509, opened 2024-03-03
Add assistant run tests

mattlindsey:
@andreibondarev I added a couple of tests and improvement(?) to assistant, but may not be what you intend.  Let me know if you can, or feel free to fix and use it.

### DIFF
--- a/lib/langchain/assistants/assistant.rb
+++ b/lib/langchain/assistants/assistant.rb
@@ -86,7 +86,7 @@ module Langchain
           end
         when "tool"
           # Run it!
-          response = chat_with_llm(tool_choice: "tool")
+          response = chat_with_llm(tool_choice: "none")
           running = true
 
           if response.tool_calls
@@ -128,9 +128,9 @@ module Langchain
     def chat_with_llm(tool_choice: "auto")
       params = {messages: thread.openai_messages}
 
-      if tools.any? && (tool_choice == "auto" || tool_choice == "tool")
+      if tools.any?
         params[:tools] = tools.map(&:to_openai_tools).flatten
-        params[:tool_choice] = tool_choice
+        params[:tool_choice] = tool_choice unless tool_choice == "none"
       end
 
       llm.chat(**params)
--- a/lib/langchain/assistants/assistant.rb
+++ b/lib/langchain/assistants/assistant.rb
@@ -73,7 +73,7 @@ module Langchain
           end
         when "user"
           # Run it!
-          response = chat_with_llm
+          response = chat_with_llm(tool_choice: "auto")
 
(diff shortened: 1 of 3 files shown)


mattlindsey:
Something a little weird is going on, since I added an expectation in the last commit (commented out), and it gives this error if you include it:

```bash
Failures:

  1) Langchain::Assistant#run when auto_tool_execution is true runs the assistant and automatically executes tool calls
     Failure/Error: expect(subject.thread.messages[-1].content).to eq("Here is the result: 4")

       expected: #<Encoding:UTF-8> "Here is the result: 4"
            got: #<Encoding:US-ASCII> ""

       (compared using ==)
     # ./spec/langchain/assistants/assistant_spec.rb:93:in `block (4 levels) in <top (required)>'
     # ./spec/support/vcr.rb:22:in `block (3 levels) in <top (required)>'
     # /Users/mattlindsey/.rvm/gems/ruby-3.2.1/gems/vcr-6.2.0/lib/vcr.rb:274:in `turned_off'
     # ./spec/support/vcr.rb:22:in `block (2 levels) in <top (required)>'
```

mattlindsey:
Also I noticed I was probably not using the `tool_choice` ('auto', 'none', or supplied function name (I think)) correctly and tried to fix it, but I'm not sure it's quite correct.

lib/langchain/assistants/assistant.rb
```diff
@@ -125,13 +125,12 @@ def submit_tool_output(tool_call_id:, output:)
...
     # @return [Langchain::LLM::BaseResponse] The LLM response object
-    def chat_with_llm
+    def chat_with_llm(tool_choice: "auto")
       params = {messages: thread.openai_messages}
 
       if tools.any?
         params[:tools] = tools.map(&:to_openai_tools).flatten
-        # TODO: Not sure that tool_choice should always be "auto"; Maybe we can let the user toggle it.
-        params[:tool_choice] = "auto"
+        params[:tool_choice] = tool_choice unless tool_choice == "none"
```
andreibondarev, line 133: I don't quite understand, what exactly does this fix?

mattlindsey: Intention (no current requirement) is to allow running just a specific tool, and adding the param to `chat_with_llm` 'fixed' testing of different scenarios with different stubs (I think).
It may not work as-is, however, and should have a separate test for running a single tool and maybe an example in the readme.

andreibondarev: @mattlindsey The rule of thumb is that if you need to change implementation in order to make the tests pass it's either 1) the implementation is wrong or 2) your tests are poorly written.

PR #510, opened 2024-03-03
Added remove_texts method on qvrant vectorsearch

gregogalante:


lib/langchain/vectorsearch/qdrant.rb
```diff
@@ -64,6 +64,16 @@ def update_texts(texts:, ids:)
       add_texts(texts: texts, ids: ids)
     end
 
+    # Remove a list of texts from the index
+    # @param ids [Array<Integer>] The ids to remove
+    # @return [Hash] The response from the server
+    def remove_texts(ids:)
```
andreibondarev, line 70: Hmm... `remove_texts()` or should we call it `delete_texts()`?

gregogalante: I think remove is better because it is the opposite of "add".

However we can change it with delete. What do you prefer?

andreibondarev:
@gregogalante We should add this method to other vectorsearch classes.

gregogalante:
@andreibondarev I will work on this in next weeks. After the update for all vector store I think we can add a **destroy_from_vectorsearch** hook on langchainrb_rails gem like what i did on my fork: https://github.com/gregogalante/langchainrb_rails/blob/main/lib/langchainrb_rails/active_record/hooks.rb

andreibondarev:
> @andreibondarev I will work on this in next weeks. After the update for all vector store I think we can add a **destroy_from_vectorsearch** hook on langchainrb_rails gem like what i did on my fork: https://github.com/gregogalante/langchainrb_rails/blob/main/lib/langchainrb_rails/active_record/hooks.rb

Nice! I like that!

Issue #506, opened 2024-02-29
Parameter 'dimensions' are not using anymore  for embbedings.


emezac:
Test:
curl https://api.openai.com/v1/embeddings \                                                                
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer AP”I_KEY \
  -d '{
    "input": "Your text string goes here", \
    "model": "text-embedding-ada-002”,\
  }'

Please check : https://platform.openai.com/docs/api-reference/chat/create 

My quick fix over : lib/langchain/llm/openai.rb:
 ```ruby
    def embed(
      text:,
      model: defaults[:embeddings_model_name],
      encoding_format: nil,
      user: nil,
      dimensions: nil # if we expect it to be explicitly passed
    )
      raise ArgumentError.new("text argument is required") if text.empty?
      raise ArgumentError.new("model argument is required") if model.empty?
      raise ArgumentError.new("encoding_format must be either float or base64") if encoding_format && %w[float base64].include?(encoding_format)

      # Warning about the obsolete `dimensions` parameter
      if dimensions
        warn "Warning: The 'dimensions' parameter is obsolete and no longer used. The embedding dimension will automatically be determined based on the selected model."
      end

      parameters = {
        input: text,
        model: model,
        # The `dimensions` parameter is no longer included here
      }
      parameters[:encoding_format] = encoding_format if encoding_format
      parameters[:user] = user if user
```
```

andreibondarev:
@emezac I think [here](https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/llm/openai.rb#L52C1-L66) what we should do is:
1. Add the `dimensions:` param back in:
```ruby
def embed(
  ...
  dimensions: nil
)
```
2. Use either what's passed in (an override) or use the defaults
```ruby
      parameters = {
        input: text,
        model: model,
        dimensions: dimensions || default_dimension
      }
```

```

emezac:
Yeap that is a better approach.

andreibondarev:
@emezac Would you be able to open a PR please? Please make sure to write tests as well.

@prdanelli Thought on this approach?

emezac:
Yes I will

spikex:
The `dimension` param now causes a error for the _text-embedding-ada-002_ model, however **can** be used with _text-embedding-3-small_ and _text-embedding-3-large_.

I would suggest:

- Making the `dimension` parameter nil by default, and just letting the API use its default.
[...]

PR #511, opened 2024-03-05
Fix dimensions params

emezac:
Fix dimensions params.

This fix is for: https://github.com/andreibondarev/langchainrb/issues/506.

Is related with the latest changes of https://openai.com/blog/new-embedding-models-and-api-updates

Issue **Before:**

Captura de pantalla 2024-03-05 a la(s) 9 35 45 a m

After: 
Captura de pantalla 2024-03-05 a la(s) 9 47 12 a m


emezac:
@andreibondarev please take a look at my implementation and the proper test specs that I added.

lib/langchain/embeddings.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Langchain
+  module Config
+    EMBEDDING_SIZES = {
```
andreibondarev, line 5: Do we intend to put ALL of the embedding sizes across the different LLM providers in 1 constant?

emezac: I wasn't too sure about this constant and I don't know what OpenAI's evolution strategy is, but I intended to be able to have these constants available to the class if needed, as in the case of the test specification. I think this OpenAI's approach is only for embeddings

emezac: Everybody is having queries/concerns about this changes -> : https://community.openai.com/t/how-to-deal-with-different-vector-dimensions-for-embeddings-and-search-with-pgvector/602141

emezac: I added a screenshot of the current behavior and the fixed behavior.

emezac: I ran all my tests yesterday  night and all passed, not sure why currently some are broken now

andreibondarev:
@emezac Would you please take a look at the spec failures? They look like actual bugs.

emezac:
> @emezac Would you please take a look at the spec failures? They look like actual bugs.

Yeap I will take care of it.

lib/langchain/llm/openai.rb
```diff
@@ -17,7 +17,7 @@ class OpenAI < Base
       n: 1,
       temperature: 0.0,
       chat_completion_model_name: "gpt-3.5-turbo",
-      embeddings_model_name: "text-embedding-ada-002"
+      embeddings_model_name: "text-embedding-3-small"
```
andreibondarev, line 20: @emezac I modified the default model per this suggestion: https://github.com/andreibondarev/langchainrb/issues/506#issuecomment-1992709067

emezac: Ok I agree

emezac:
Sorry @andreibondarev  I was a little busy implementing a RAG application using your Gem. I was creating the implementation of the **ask** method but at the row level, I mean **ask method** at the instance level, interesting. Thanks!

PR #515, opened 2024-03-10
Added remove_texts method on chroma vectorsearch

gregogalante:


lib/langchain/vectorsearch/base.rb
```diff
@@ -167,12 +167,13 @@ def generate_hyde_prompt(question:)
     #
     # @param question [String] User's question
     # @param context [String] The context to synthesize the answer from
+    # @param prompt_template_path [String] The path to the prompt template
     # @return [String] Prompt
-    def generate_rag_prompt(question:, context:)
+    def generate_rag_prompt(question:, context:, prompt_template_path: nil, **additional_params)
```
andreibondarev, line 172: @gregogalante I do like this idea of passing a custom prompt to do RAG. Is this what you were thinking with this PR?

gregogalante: Yes but I want to close remove_texts on all vector search before

andreibondarev: @gregogalante I added a [ticket](https://github.com/andreibondarev/langchainrb/issues/520) to keep track of this work.

PR #523, opened 2024-03-16
Added remove_texts on some different vectorsearch

gregogalante:


gregogalante:
@andreibondarev I think we must modify the milvus vectorsearch to receive ids on add_texts or

PR #529, opened 2024-03-18
Add Support for Claude 3 via AWS Bedrock

mtrefilek:
* Claude 3 utilizes new messaging param
* New parameters are incompatible wiht existing Bedrock Anthropic calls

https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html


lib/langchain/llm/aws_bedrock/antrhopic_messages.rb
```diff
@@ -0,0 +1,69 @@
...
+module Langchain::LLM
+  # LLM interface for Aws Bedrock Claude Messaging APIs: https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html
+  #
+  # Gem requirements:
+  #    gem 'aws-sdk-bedrockruntime', '~> 1.1'
+  #
+  # Usage:
+  #    claude3 = Langchain::LLM::AwsBedrock::AnthropicMessages.new(llm_options: {})
+  #
+  class AwsBedrock::AnthropicMessages < AwsBedrock
```
andreibondarev, line 12: Is it possible to re-use the existing [AwsBedrock](https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/llm/aws_bedrock.rb) class?

andreibondarev: @mtrefilek I think combining it under the same [AwsBedrock](https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/llm/aws_bedrock.rb) class makes sense.

You'd basically add:
```ruby
  class AwsBedrock < Base
    DEFAULTS = {
      completion_model_name: "[anthropic.claude-v2](anthropic.claude-3-sonnet-20240229-v1:0)",
```

And then the chat method only works with that model from my understanding.

mtrefilek: @andreibondarev Good call, It actually looks like all Claude Models support this new messaging inference request. I updated the default model and forced chat to only allow messages. 

I will add a condition to error .complete on claude-3 models

To my understanding no. I believe the models should all be better and cheaper based on most reading I have seen.

lib/langchain/llm/aws_bedrock.rb
```diff
@@ -11,13 +11,13 @@ module Langchain::LLM
...
     DEFAULTS = {
-      completion_model_name: "anthropic.claude-v2",
+      completion_model_name: "anthropic.claude-3-sonnet-20240229-v1:0",
       embedding_model_name: "amazon.titan-embed-text-v1",
       max_tokens_to_sample: 300,
       temperature: 1,
       top_k: 250,
       top_p: 0.999,
-      stop_sequences: ["\n\nHuman:"],
+      stop_sequences: [],
```
andreibondarev, line 20: Just curious -- why this change?

lib/langchain/llm/aws_bedrock.rb
```diff
@@ -105,6 +106,26 @@ def complete(prompt:, **params)
       parse_response response
     end
 
+    def chat(messages: [], system: nil, **params)
```
andreibondarev, line 109: Instead of this `**params`, I'd like to actually be super explicit (similar to https://github.com/andreibondarev/langchainrb/blob/main/lib/langchain/llm/openai.rb#L105-L125). The reason for that is because as a developer, when you look at `**params`, you have no clue what's actually expected. For the interfaces that you control, such as this one, it makes sense to enumerate all of the different options.

lib/langchain/llm/aws_bedrock.rb
```diff
@@ -178,6 +199,19 @@ def compose_parameters_anthropic(params)
...
+    def compose_parameters_antrhopic_messaging(params)
+      default_params = @defaults.merge(params)
+
+      {
+        max_tokens: default_params[:max_tokens],
+        temperature: default_params[:temperature],
+        top_k: default_params[:top_k],
+        top_p: default_params[:top_p],
+        stop_sequences: default_params[:stop_sequences],
+        anthropic_version: default_params[:anthropic_version]
```
andreibondarev, line 211: Is this an actual valid option to pass in? Isn't the version in the model name, i.e.: "anthropic.claude-3-sonnet-20240229-v1:0"?

mtrefilek:
@andreibondarev Sorry for my delay. This should be ready for another review providing we are okay locking `#chat` to only the Claude family of models and this new request syntax. (It makes sense based on the difference in calling)

I can provide some code examples with regular messaging and multimodal (image) messaging later.

lib/langchain/llm/response/anthropic_messages_response.rb
```diff
@@ -0,0 +1,49 @@
...
+  class AnthropicMessagesResponse < BaseResponse
+    def model
+      raw_response.dig("model")
+    end
+
+    def chat_completions
+      raw_response.dig("content")
+    end
+
+    def chat_completion
```
andreibondarev, line 13: Could you please see if you could reuse the `AnthropicResponse` class I added here? https://github.com/patterns-ai-core/langchainrb/pull/580/files

mtrefilek: @andreibondarev Double checking tests, but it is working in my dev enviroment!

mtrefilek:
> @mtrefilek 1 last comment and I think this can be shipped! 🚢

@andreibondarev I think that worked well!

andreibondarev:
@mtrefilek Thank you for your persistence and contribution! 👏🏻

PR #538, opened 2024-03-23
Pass source: to the Langchain::Chunk

andreibondarev:


lib/langchain/vectorsearch/base.rb
```diff
@@ -183,16 +183,21 @@ def generate_rag_prompt(question:, context:)
...
       raise ArgumentError, "Paths must be provided" if Array(paths).empty?
 
-      texts = Array(paths)
+      chunks = Array(paths)
         .flatten
         .map do |path|
-          data = Langchain::Loader.new(path, options, chunker: chunker)&.load&.chunks
-          data.map { |chunk| chunk.text }
-        end
+          options[:source] ||= path.to_s
```
jjimenez, line 189: If I pass in an array, each path might have it's own metadata / source.  I would like to see this be able to take a string or a hash in the array. I hope this suggestion helps!

```suggestion
         if path.is_a?(Hash)
           raise ArgumentError "Path hashes must have keys :path, :source" unless (path.keys & [:path, :source] == [:path, :source])
            options[:source] ||= path[:source]
            path = path[:path]
        end
        raise ArgumentError "Paths must be strings or hashes with keys :path, :source" unless path.is_a?(String)
          options[:source] ||= path.to_s
```

PR #539, opened 2024-03-23
Update mistral_ai.rb to use api_key parameter instead of ENV

max-power:


andreibondarev (approved):
@max-power Thank you for your PR!

Issue #549, opened 2024-03-28
Chunking CSV file


Eth3rnit3:
### Problem:
When attempting to parse CSV files using the gem, an error occurs due to improper handling of text chunking.

```ruby
file_path       = 'data/folder/test.csv'
abs_path        = File.expand_path(file_path)

llm = Langchain::LLM::OpenAI.new(api_key: ENV['OPEN_AI_KEY'])
db = Langchain::Vectorsearch::Chroma.new(
  url: ENV['CHROMA_DB_URL'],
  index_name: ENV.fetch('CHROMA_DB_INDEX_NAME', 'documents'),
  llm:
)
db.add_data(paths: [Langchain.root.join(abs_path)], options: { col_sep: ';' })
```

It raise following error
```shell
splits = separator.empty? ? text.chars : text.split(separator)
                                                   ^^^^^^
from /Users/david/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/baran-0.1.11/lib/baran/text_splitter.rb:21:in `chunks'
from /Users/david/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/langchainrb-0.10.2/lib/langchain/chunker/text.rb:33:in `chunks'
from /Users/david/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/langchainrb-0.10.2/lib/langchain/data.rb:26:in `chunks'
from /Users/david/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/langchainrb-0.10.2/lib/langchain/vectorsearch/base.rb:189:in `block in add_data'
from /Users/david/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/langchainrb-0.10.2/lib/langchain/vectorsearch/base.rb:188:in `map'
from /Users/david/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/langchainrb-0.10.2/lib/langchain/vectorsearch/base.rb:188:in `add_data'
from test.rb:12:in `<main>'
```

### Context:

The error arises within the Langchain::Vectorsearch::Chroma module when processing CSV files. The current implementation fails to accommodate arrays of strings, such as those generated by reading CSV files, leading to the TypeError mentioned above.

Eth3rnit3:
I can confirm that the problem has been corrected 👍 well done guys for all your hard work!

andreibondarev:
[...]

PR #552, opened 2024-04-02
Add ability to chunk CSV in either row or file

WToa:
Looking to address the issue in https://github.com/andreibondarev/langchainrb/issues/549


lib/langchain/processors/csv.rb
```diff
@@ -22,6 +33,29 @@ def parse(data)
...
+        end
+      end
+
+      def chunk_row(data)
+        ::CSV.new(data.read, col_sep: separator).map do |row|
+          row
+            .compact
+            .map(&:strip)
+            .join(separator)
+        end.join("\n\n")
```
WToa, line 53: From what I can tell, by default if a string is separated by `\n\n` then it gets chunked when it's acted on by `Langchain::Chunker::Text` Let me know if that's incorrect or you prefer a different approach

andreibondarev:
@WToa Thank you for the PR!

PR #561, opened 2024-04-05
Add file tool refactored

andreibondarev:


lib/langchain/tool/file_system/file_system.json
```diff
@@ -0,0 +1,57 @@
...
+    "type": "function",
+    "function": {
+      "name": "file_system-write_to_file",
+      "description": "File System Tool: Write content to a file",
+      "parameters": {
+        "type": "object",
+        "properties": {
+          "file_path": {
+            "type": "string",
+            "description": "Directory path to list"
```
timkleier, line 46: "Path to the file to write"

lib/langchain/tool/file_system/file_system.rb
```diff
@@ -0,0 +1,34 @@
...
+    ANNOTATIONS_PATH = Langchain.root.join("./langchain/tool/#{NAME}/#{NAME}.json").to_path
+
+    description <<~DESC
+      A tool that wraps the Ruby File class, for interacting with the file system.
+    DESC
+
+    def list_directory(directory_path:)
+      Dir.entries(directory_path)
+    rescue Errno::ENOENT
+      "No such directory: #{directory_path}"
```
timkleier, line 21: Do we want this message just returned, or logged, or both?

lib/langchain/tool/file_system/file_system.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+
+    def read_file(file_path:)
+      File.read(file_path)
+    rescue Errno::ENOENT
+      "No such file: #{file_path}"
+    end
+
+    def write_to_file(file_path:, content:)
+      File.write(file_path, content)
```
timkleier, line 31: Do we want to handle error scenarios here, like permission denied? And should that apply to the other methods as well?

lib/langchain/tool/file_system/file_system.rb
andreibondarev: Definitely returned -- it communicates to the LLM what the issue is.

I've been thinking about a clean logging system when the tools are executed, something like:
image

lib/langchain/tool/file_system/file_system.rb
andreibondarev: Yeah, I just haven't been able to generate all the different error scenarios. Idk if it's all `Errno::ENOENT` that's raised.

PR #574, opened 2024-04-12
Ensure that default dimensions passed to LLM instance are respected by embed

prdanelli:
Related to: https://github.com/patterns-ai-core/langchainrb/issues/572

We need to set the embed dimensions in the LLM instance, otherwise we risk having one the Vector DB config set to a different dimension size to the dimensions returned by the embed method. 

Allow `dimensions` to be manually set in the `embed` method, but respect them if they are set in the LLM.

### DIFF
--- a/lib/langchain/llm/openai.rb
+++ b/lib/langchain/llm/openai.rb
@@ -185,7 +185,7 @@ module Langchain::LLM
     end
 
     def default_dimension
-      @defaults[:dimension] || EMBEDDING_SIZES.fetch(defaults[:embeddings_model_name])
+      @defaults[:dimensions] || EMBEDDING_SIZES.fetch(defaults[:embeddings_model_name])
     end
 
     private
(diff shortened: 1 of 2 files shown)


andreibondarev:
@prdanelli I think this parameter needs to be changed to plural: https://github.com/patterns-ai-core/langchainrb/blob/main/lib/langchain/llm/openai.rb#L188.

In fact -- I think we need to make that global change everywhere: plural ("dimensions") instead or "dimension".

prdanelli:
@andreibondarev absolutely agree about the item you linked - i've pushed that fix. The rest should probably be updated in another PR as it might be a breaking change.

andreibondarev:
@prdanelli Once the specs are fixed -- let's ship this PR.

Ticket for the future: https://github.com/patterns-ai-core/langchainrb/issues/575

prdanelli:
Just pushed a fix... I missed one of the dimensions references in that file.

prdanelli:
@andreibondarev all ✅ now.

prdanelli:
Thanks for your help getting this sorted Andrei. 🎖️

andreibondarev:
Thank you for your fix!

PR #579, opened 2024-04-15
function calling with structured output using EasyTalk gem.

sergiobayona:
Define the JSON Schema inside a Ruby class using the EasyTalk gem. Then generate a function with the schema and pass it to the OpenAI's `chat` method via the `tools` argument. Run the call and expect valid JSON output.


andreibondarev:
@sergiobayona Thank you for your effort here. I think the current DSL, enabled by the `Langchain::ToolDefinition` module is sufficient.

PR #583, opened 2024-04-19
Translate max_tokens_to_sample for OpenAI

scottcmoore:
While attempting to use the semantic chunker with OpenAI, I ran into this issue:
```
[2] pry(main)> url = "https://www.scottcmoore.com"
data = Langchain::Loader.load(url).value
llm = LlmFactory.create!("OpenAI")
Langchain::Chunker::Semantic.new(
  data,
  llm: llm
DEPRECATED: `Langchain::LLM::OpenAI#complete` is deprecated, and will be removed in the next major version. Use `Langchain::LLM::OpenAI#chat` instead.
ArgumentError: unknown keyword: :max_tokens_to_sample
```

OpenAI expects this parameter as 'max_tokens', while other providers appear to expect 'max_tokens_to_sample'. This PR updates the OpenAI chat method to accept either version of the parameter. 
Let me know what you think of this approach, thanks!

### DIFF
--- a/lib/langchain/llm/openai.rb
+++ b/lib/langchain/llm/openai.rb
@@ -112,6 +112,7 @@ module Langchain::LLM
       logprobs: nil,
       top_logprobs: nil,
       max_tokens: nil,
+      max_tokens_to_sample: nil,
       n: defaults[:n],
       presence_penalty: nil,
       response_format: nil,
@@ -128,6 +129,8 @@ module Langchain::LLM
       raise ArgumentError.new("messages argument is required") if messages.empty?
       raise ArgumentError.new("model argument is required") if model.empty?
       raise ArgumentError.new("'tool_choice' is only allowed when 'tools' are specified.") if tool_choice && tools.empty?
+
+      max_tokens ||= max_tokens_to_sample
 
       parameters = {
         messages: messages,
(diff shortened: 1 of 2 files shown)


andreibondarev:
@scottcmoore Which LLM are you using if you don't mind me asking? If instead of this we dropped the `max_tokens_to_sample: 50000` from this [line](https://github.com/patterns-ai-core/langchainrb/blob/0.11.3/lib/langchain/chunker/semantic.rb#L29), would it still work for you? 

The way I'm thinking about this is that OpenAI sets the "gold-standard" in terms of API interfaces and we adapt all other LLM provider interfaces to match it.

In practice this looks like the following... Every `Langchain::LLM::Base` subclass implements the following interface:
```ruby
def chat(
  messages: [],
  model: defaults[:chat_completion_model_name],
  max_tokens: nil
  ...
)
```
But the underlying calls, depending on the LLM class, may need to remap to their respective `maxTokens` or `max_tokens_to_sample`, etc.

I won't claim that this is implemented everywhere with 100% accuracy that this is where I'd like it to converge towards.

scottcmoore:
@andreibondarev I was using `gpt-3.5-turbo` when testing this.
If the `max_tokens_to_sample: 50000` were dropped I think it would work fine for our use case. 

Just to make sure, am I understanding correctly that you'd like to keep the OpenAI provider interface as it is, and update the others to accept max tokens as a param to their chat methods, and do any necessary translation of parameter names in those providers?

PR #591, opened 2024-04-25
Minor database tool improvements

mattlindsey:
Improves performance of db tool with 2 updates:

- list_tables returns a simple comma-separated list instead of an array of symbols since llms were interpreting the `:` at the start of each symbol as part of the table name.
- execute returns e.message when there's an exception, which helps llms retry when a query doesn't work
Tested with both openai and google gemini PR.

Fixes https://github.com/patterns-ai-core/langchainrb/issues/432


mattlindsey:
Fixed by #743

PR #593, opened 2024-04-26
`Langchain::LLM::Base#chat()` accepts a new `response_schema:` parameter to force the response to adhere to JSON schema

andreibondarev:
This PR introduces a new parameter `response_schema:` to the LLM `chat()` methods that, using Function Calling, forces the response to adhere to a specific schema. Example usage:
```ruby
json_schema = {
  type: "object",
  properties: {
    name: {
      type: "string",
      description: "Persons name"
    },
    age: {
      type: "number",
      description: "Persons age"
    },
    interests: {
      type: "array",
      items: {
        type: "object",
        properties: {
          interest: {
            type: "string",
            description: "A topic of interest"
          },
          levelOfInterest: {
            type: "number",
            description: "A value between 0 and 100 of how interested the person is in this interest"
          }
        },
        required: ["interest", "levelOfInterest"],
        additionalProperties: false
      },
      minItems: 1,
      maxItems: 3,
      description: "A list of the person's interests"
    }
  },
  required: ["name", "age", "interests"],
  additionalProperties: false
}

response = llm.chat messages: [{ role: :user, content: "Extract Jason is 25 years, and likes to play soccer"}], response_schema: json_schema

response.response_schema
#=> {"name"=>"Jason", "age"=>25, "interests"=>[{"interest"=>"soccer", "levelOfInterest"=>80}]}
```


sergiobayona:
This is good and it is a move in the right direction but there are some key points that are missing:

1. Generating the json_schema only is not enough. The JSON Schema specification provides validation vocabulary. For example `format`, `min_length`, `enum`, `default`, etc. Check out https://json-schema.org/draft/2020-12/json-schema-validation. EasyTalk does this with a `validate()` method. Ensuring that the returned data validates correctly against the schema is probably the most important thing the Instructor library does.
2. Retry logic. If the returned data is invalid, it does a retry to indicate to the LLM that the payload was invalid and coerce it to correct return the right data.
3. Support for multiple objects. This is basically "parallel function calling". We were talking about this yesterday. If you run a function call asking "What is the weather in Boston, Los Angeles and Miami?" the LLM will return 3 items (objects) in the payload. The code does not account for that scenario.

There are other more advanced features like streaming and threading or Ruby fibers so that you can process calls in parallel (different from parallel function calling) etc that I want to add Instructor-rb. Those scaling features are the money makers and the reason why Instructor has been successful.

lib/langchain/llm/response/openai_response.rb
```diff
@@ -28,6 +28,13 @@ def tool_calls
       chat_completions&.dig(0, "message", "tool_calls")
     end
 
+    # @return [Hash] JSON schema structured response
+    def response_schema
+      if tool_calls
+        JSON.parse(tool_calls.first.dig("function", "arguments"))
```
sergiobayona, line 34: you have a module called `Langchain::LLM::OpenAIResponse` you should use it here.

sergiobayona: what happens when the LLM sends back an array of objects?

andreibondarev: It doesn't work with parallel function-calling right now if that's what you're asking.

andreibondarev:
> This is good and it is a move in the right direction but there are some key points that are missing:
> 
> 1. Generating the json_schema only is not enough. The JSON Schema specification provides validation vocabulary. For example `format`, `min_length`, `enum`, `default`, etc. Check out https://json-schema.org/draft/2020-12/json-schema-validation. EasyTalk does this with a `validate()` method. Ensuring that the returned data validates correctly against the schema is probably the most important thing the Instructor library does.
> 2. Retry logic. If the returned data is invalid, it does a retry to indicate to the LLM that the payload was invalid and coerce it to correct return the right data.
> 3. Support for multiple objects. This is basically "parallel function calling". We were talking about this yesterday. If you run a function call asking "What is the weather in Boston, Los Angeles and Miami?" the LLM will return 3 items (objects) in the payload. The code does not account for that scenario.
> 
> There are other more advanced features like streaming and threading or Ruby fibers so that you can process calls in parallel (different from parallel function calling) etc that I want to add Instructor-rb. Those scaling features are the money makers and the reason why Instructor has been successful.

1. Using EasyTalk, can I do something like `UserDetail.new(raw_json).validate()`?
2. Does instructor always retry or can you "do it once, fail and raise an error"?
3. Correct, parallel function calling is not supported yet.

palladius:
QQ. Yesterday I learnt at RubyDay of a recent Data class (since 3.2) for immutable hashes (https://www.shakacode.com/blog/ruby-3-2-adds-a-new-data-class/). This seems like a perfect fit - not much for the json_schema but to force the result to fit into a Data which you might construct based on the `json_schema`. 
Just thinking out loud here - would it make sense? Feel free to shoot me down if I said sth stupid - which is likely.

[ I say this because I'm using it to accept different Gemini responses (good vs error response) by forcing it into two wlel-known, different schema. ]

andreibondarev:
> QQ. Yesterday I learnt at RubyDay of a recent Data class (since 3.2) for immutable hashes (https://www.shakacode.com/blog/ruby-3-2-adds-a-new-data-class/). This seems like a perfect fit - not much for the json_schema but to force the result to fit into a Data which you might construct based on the `json_schema`. Just thinking out loud here - would it make sense? Feel free to shoot me down if I said sth stupid - which is likely.
> 
> [ I say this because I'm using it to accept different Gemini responses (good vs error response) by forcing it into two wlel-known, different schema. ]

@palladius I was thinking of returning a simple hash so that the user could do whatever they want with it?

andreibondarev:
Since LLMs are starting to support JSON schema response natively -- there's no need to try and do this "manually". We merged this [PR](https://github.com/patterns-ai-core/langchainrb/pull/766) instead.

PR #595, opened 2024-04-26
OpenAI's latest model is reflected in TokenValidator

### LINKED ISSUE #594: OpenAI new models(`gpt-4-turbo`, `gpt-3.5-turbo-0125`) can not use with `undefined method `-' for nil (NoMethodError)`
# Description

Errors occur when using the latest models `gpt-4-turbo` and `gpt-3.5-turbo-0125`.

```
>  llm = Langchain::LLM::OpenAI.new(
     api_key: ENV.fetch('OPENAI_API_KEY', nil),
     default_options: {
       chat_completion_model_name: 'gpt-4-turbo'
   })
> llm.chat(messages: [{ role: 'user', content: 'hello'}])
Warning: gpt-4 may update over time. Returning num tokens assuming gpt-4-0613.
/Users/kokuyou/.asdf/installs/ruby/3.3.1/lib/ruby/gems/3.3.0/gems/langchainrb-0.8.2/lib/langchain/utils/token_length/base_validator.rb:22:in `validate_max_tokens!': undefined method `-' for nil (NoMethodError)

          leftover_tokens = token_limit(model_name) - text_token_length
                                                    ^
	from /Users/kokuyou/.asdf/installs/ruby/3.3.1/lib/ruby/gems/3.3.0/gems/langchainrb-0.8.2/lib/langchain/utils/token_length/openai_validator.rb:75:in `validate_max_tokens!'
	from /Users/kokuyou/.asdf/installs/ruby/3.3.1/lib/ruby/gems/3.3.0/gems/langchainrb-0.8.2/lib/langchain/llm/openai.rb:242:in `validate_max_tokens'
	from /Users/kokuyou/.asdf/installs/ruby/3.3.1/lib/ruby/gems/3.3.0/gems/langchainrb-0.8.2/lib/langchain/llm/openai.rb:134:in `chat'
	from (irb):19:in `<main>'
	from <internal:kernel>:187:in `loop'
	from bin/console:155:in `<main>'
```

My guess is that the contents of Langchain::Utils::TokenLength have not been updated.

kokuyouwind:
# Description

fix #594 

Based on [OpenAI's most recent list of models](Based on OpenAI's most recent list of models, the missing models were added and updates to existing models were reflected.), the missing models were added and updates to existing models were reflected.
I checked the specs for OpenAIValidator and I don't think it needs to be updated.


lib/langchain/utils/token_length/openai_validator.rb
```diff
@@ -28,10 +28,11 @@ class OpenAIValidator < BaseValidator
           "text-embedding-3-large" => 8191,
           "text-embedding-3-small" => 8191,
           "text-embedding-ada-002" => 8191,
-          "gpt-3.5-turbo" => 4096,
+          "gpt-3.5-turbo" => 16385,
```
kokuyouwind, line 31: :memo:
gpt-3.5-turbo is now aliased to gpt-3.5-turbo-0125, so I aligned it there.

andreibondarev:
@kokuyouwind Do you find there's value in this token length validation? I've been thinking of disabling it by default.

kokuyouwind:
@andreibondarev 

> @kokuyouwind Do you find there's value in this token length validation? I've been thinking of disabling it by default.

I have created this PR with respect to the existing implementation, but I too think this validation is unnecessary.
If there is a problem with the number of tokens, the OpenAI API will return an error, and I don't think there is a particular need to save the number of API runs.

PR #596, opened 2024-04-28
Add OpenAI-compatible Groq

mattlindsey:
TODO:
- [x] chat method works with Groq models
- [x] Add tests
- [x] Add usage to README
- [x] Confirm that Assistant with tools/function-calling works ok with different models


README.md
```diff
@@ -105,6 +106,32 @@ You can use any other LLM by invoking the same interface:
 llm = Langchain::LLM::GooglePalm.new(api_key: ENV["GOOGLE_PALM_API_KEY"], default_options: { ... })
 ```
 
+#### Groq OpenAI Compatible
+
+Add `gem "ruby-openai", "~> 7.0"` to your Gemfile.
+
+```ruby
+llm = Langchain::LLM::GroqOpenAI.new(api_key: ENV["GROQ_API_KEY"]) # Defaults to Meta Llama 3
```
andreibondarev, line 114: I would prefer to call it: `Langchain::LLM::Groq`. I find `GroqOpenAI` rather confusing as it sounds like the OpenAI models are available on Groq, which they're not.

lib/langchain/llm/groq_open_ai.rb
```diff
@@ -0,0 +1,25 @@
...
+module Langchain::LLM
+  # LLM interface for Groq OpenAI compatible services
+  #
+  # Usage:
+  #    groq = Langchain::LLM::GroqOpenAi.new(
+  #      api_key: ENV["GROQ_API_KEY"],
+  #      llm_options: {},
+  #      default_options: {}
+  #    )
+  class GroqOpenAi < OpenAI
```
andreibondarev, line 12: I would build a standalone integration to Groq directly instead of subclassing OpenAI. I understand that it works but it only works because Groq is trying to mimic the OpenAI interface.

andreibondarev: @mattlindsey Not use `ruby-openai` gem and not subclass for Groq.

PR #599, opened 2024-04-29
Add support for .xls (Excel 1997-2004)

bricolage:
Langchain currently has support for XLSX/XLSM but not for the prior XLS format, aka Excel 1997-2004.

Roo has a companion gem roo-xls just for this purpose. This loader and specs are nearly identical to the existing Xlsx version.


andreibondarev:
@bricolage Thank you for the PR!

PR #600, opened 2024-04-30
Introduce UnifiedParameters

codenamev:
This is an initial effort to unify the API signature for all the LLMs' `chat` method.  This follows [OpenRouter's philosophy](https://openrouter.ai/docs#requests) of "normalizes the schema across models and providers so you only need to learn one".  Each LLM will need to implement their own adaptations of the base-unified parameter list, and follow some simple rules:

1. All `LLM` subclasses will now use an instance of `UnifiedParameters` and can access it via `chat_parameters`
2. All `LLM` subclasses can customize the schema used by the unified parameters in their initializer
3. If the chosen model doesn't support a request parameter (such as logit_bias in non-OpenAI models, or top_k for OpenAI), then the parameter will be ignored. The rest will be forwarded to the underlying model API.

This is also an initial effort to set up the foundation for a [Parameters API](https://openrouter.ai/docs#parameters).  The idea being that each model can use their customized UnifiedParameters as an interface for querying what those parameters are.

I'd love some feedback on this before I go spreading it to every other LLM class 🥺


lib/langchain/llm/openai.rb
```diff
@@ -40,6 +40,16 @@ def initialize(api_key:, llm_options: {}, default_options: {})
       @client = ::OpenAI::Client.new(access_token: api_key, **llm_options)
 
       @defaults = DEFAULTS.merge(default_options)
+      chat_parameters.update(
+        model: {default: "gpt-3.5-turbo"},
```
andreibondarev, line 44: Should we be using `@defaults[:chat_completion_model_name]` here?

lib/langchain/llm/openai.rb
```diff
@@ -104,52 +114,14 @@ def complete(prompt:, **params)
...
-      top_p: nil,
-      tools: [],
-      tool_choice: nil,
-      user: nil,
-      &block
-    )
-      raise ArgumentError.new("messages argument is required") if messages.empty?
-      raise ArgumentError.new("model argument is required") if model.empty?
-      raise ArgumentError.new("'tool_choice' is only allowed when 'tools' are specified.") if tool_choice && tools.empty?
+    def chat(params = {}, &block)
```
andreibondarev, line 117: I tried overriding the `model:` when calling the `chat()` method but it didn't work:
```ruby
irb(main):009> openai.chat messages:[{role:"user", content:"hello!"}], model: "gpt-4"
=>
#<Langchain::LLM::OpenAIResponse:0x0000000136817dd8
 @model=nil,
 @raw_response=
  {"id"=>"chatcmpl-9K2yrYv6ayNahExEymLN6ncBFSlqg",
   "object"=>"chat.completion",
   "created"=>1714565045,
   "model"=>"gpt-3.5-turbo-0125",
```

andreibondarev:
@codenamev I should finish my review by the end of today.

lib/langchain/llm/openai.rb
```diff
@@ -104,52 +114,14 @@ def complete(prompt:, **params)
...
-      response_format: nil,
-      seed: nil,
-      stop: nil,
-      stream: nil,
-      temperature: defaults[:temperature],
-      top_p: nil,
-      tools: [],
-      tool_choice: nil,
-      user: nil,
-      &block
```
andreibondarev, line 126: @codenamev You didn't like how all of the params were listed out? We had that previously because I felt like it made the interface clearer since it's an interface that we control.

andreibondarev:
@codenamev I'm thinking that maybe this PR targets a feature branch `unified-parameters` and until we've converted all of the other LLMs we'd hold back rolling this out. What do you think?

lib/langchain/llm/openai.rb
codenamev: I was torn here, but allowing these inherited classes to change the method signature with custom keyword arguments complicates things quite a lot and is adding confusion on when to use what.  One thing we could do is change this to accept a `UnifiedParameters` instance (instead of a Hash as it is) and then pass that along to the `chat_parameters` the same way (and just have it merge).  Doing that would allow us to continue documenting and inspect what is acceptable criteria (similar to how GraphQL mutations can accept input types).

codenamev:
> @codenamev I'm thinking that maybe this PR targets a feature branch `unified-parameters` and until we've converted all of the other LLMs we'd hold back rolling this out. What do you think?

Is there a branch named that, or are you just saying we should do all or nothing?  Heh, I meant to submit a Draft PR here just showcasing how it would work with the most common LLM (OpenAI).  If we can agree on the abstraction and new parameter flow, I'd be happy to run through all the existing LLMs and update them to mimic the same behavior :D

lib/langchain/llm/openai.rb
codenamev: Just pushed a fix.  Was overly caching calls to the UnifiedParameters#to_params method.

lib/langchain/llm/openai.rb
andreibondarev: Yeah, the idea is just to provide some sensible defaults to allow people to get decent results quickly without thinking much.

lib/langchain/llm/openai.rb
andreibondarev: What would the DX look like then? Something along the lines of:
```ruby
uniparams = Langchain::LLM::UnifiedParameters.new schema: {}, parameters: params
openai.chat parameters: uniparams
```
?

andreibondarev:
> > @codenamev I'm thinking that maybe this PR targets a feature branch `unified-parameters` and until we've converted all of the other LLMs we'd hold back rolling this out. What do you think?
> 
> Is there a branch named that, or are you just saying we should do all or nothing? Heh, I meant to submit a Draft PR here just showcasing how it would work with the most common LLM (OpenAI). If we can agree on the abstraction and new parameter flow, I'd be happy to run through all the existing LLMs and update them to mimic the same behavior :D

Yeah, I just meant that we should do all or nothing. I can cut a new `origin/unified-parameters` branch if you think it would be helpful.

lib/langchain/llm/unified_parameters.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+module Langchain::LLM
+  class UnifiedParameters
```
andreibondarev, line 4: @codenamev Think we'd jam all of the parameters for `chat()`, `embed()` and `complete()` under `UnifiedParameters` or we'd introduce individual classes for `UnifiedChatParameters`, `UnifiedEmbedParameters`, `UnifiedCompleteParameters`?

lib/langchain/llm/openai.rb
codenamev: Just to re-iterate, the goal of these changes is that `Langchain::LLM::Base` is the source of truth for the unified method signature we want applied to all children.  Preserving this idea, the LLM sub-classes will be responsible for:

1. Adhering to the unified method signature of its' parent
2. Managing any amendments to the method signature it wants to handle

The way this currently is implemented, each sub-class uses a `chat_params` instance from `LLM::Base` and modifies it [in its' initializer](https://github.com/patterns-ai-core/langchainrb/pull/600/files#diff-e9b54ad7b6ef063bedb31fba426cea87e7b816d4eeaa65a4ffb07929e6e58646R43-R52).

Following `3` of OpenRouter's principles here:

> If the chosen model doesn't support a request parameter, then the parameter will be ignored. The rest will be forwarded to the underlying model API.

The callers should be able to pass _any_ parameters to these sub-classes and the model itself will know how to handle them for their respective API calls.  For this reason, this is why I modified the arguments of `chat` for OpenAI to accept a hash rather than explicit keyword arguments.  Under the hood, the _model_ forces the unified schema, ignores any fields, re-maps (aliases) any fields it needs re-named, and allows any additional fields the model accepts.

In the future, we can add validations and parameter inquiries, but this is mostly a small step backward (obfuscating the method signature) in favor of applying the unified signature with backwards compatibility.  I'll be adding further documentation for what these parameters are that I think will solve some of your worries here 😄

lib/langchain/llm/unified_parameters.rb
codenamev: This makes a lot of sense.  I originally had a `LLM::Parameters::Chat`, but removed it because it was just the schema hash.  Let me refactor it back.  It will certainly make documenting things easier.

lib/langchain/llm/openai.rb
andreibondarev: @codenamev Okay, I love it, and I'm on board 🚢 🚀

andreibondarev: (A proper docs website is desperately needed) but we'd probably still want to catalogue all of the possible values that could be passed under the `params={}` hash [here](https://github.com/patterns-ai-core/langchainrb/pull/600/files#diff-e9b54ad7b6ef063bedb31fba426cea87e7b816d4eeaa65a4ffb07929e6e58646R114-R116) so that it shows up in the [rubydoc](https://rubydoc.info/gems/langchainrb/Langchain/LLM/OpenAI#chat-instance_method) as well.

lib/langchain/llm/base.rb
```diff
@@ -70,5 +70,15 @@ def embed(...)
     def summarize(...)
       raise NotImplementedError, "#{self.class.name} does not support summarization"
     end
+
+    #
+    # Returns an instance of Langchain::LLM::Parameters::Chat
+    #
+    def chat_parameters(params = {})
+      @chat_parameters ||= Langchain::LLM::Parameters::Chat.new(
```
andreibondarev, line 78: I think the chat params are being cached:
```ruby
irb(main):017> llm = Langchain::LLM::Anthropic.new api_key: ENV['ANTHROPIC_API_KEY']
=>
#<Langchain::LLM::Anthropic:0x000000012e51b010
...
irb(main):018> llm.chat messages: [{role: :user, content: "hey"}], model: "gpt-3.5-turbo"
=>
#<Langchain::LLM::AnthropicResponse:0x000000012e911160

irb(main):020> llm.chat messages: [{role: :user, content: "hey"}]
=>
#<Langchain::LLM::AnthropicResponse:0x000000012e6d2390
 @model=nil,
 @raw_response={"type"=>"error", "error"=>{"type"=>"not_found_error", "message"=>"model: gpt-3.5-turbo"}}>
```

andreibondarev:
@codenamev I just noticed that the response from Ollama comes back serialized:
```ruby
irb(main):005> llm.chat messages: [{role: "user", content: "Hey"}], stop: ["ello"]

=>
#<Langchain::LLM::OllamaResponse:0x00000001299d6118
 @model="llama3",
 @prompt_tokens=nil,
 @raw_response=
  "{\"model\":\"llama3\",\"created_at\":\"2024-05-10T16:54:49.041769Z\",\"message\":{\"role\":\"assistant\",\"content\":\"Hey\"},\"done\":false}\n{\"model\":\"llama3\",\"created_at\":\"2024-05-10T16:54:49.062254Z\",\"message\":{\"role\":\"assistant\",\"content\":\"!\"},\"done\":false}\n{\"model\":\"llama3\",\"created_at\":\"2024-05-10T16:54:49.082952Z\",\"message\":{\"role\":\"assistant\",\"content\":\" How\"},\"done\":false}\n{\"model\":\"llama3\",\"created_at\":\"2024-05-10T16:54:49.103971Z\",\"message\":{\"role\":\"assistant\",\"content\":\"'s\"},\"done\":false}\n{\"model\":\"ll
  ...
```

andreibondarev:
@codenamev Tremendous effort, great job!

drnic:
Huge effort!

PR #602, opened 2024-05-01
Fix weather tool exceptions and units parameter

### LINKED ISSUE #175: Weather questions sometimes blows up on certain city names
ChainOfThought agent usually provides a good city name to the Weather tool, but can blow up (or be tricked into blowing up) on city names.

First example works, second blows up on same city:

image

mattlindsey:
Fixes #175 and replaces PR #193.

- Catching the exception is still necessary since api errors still occur, and this allows llm to retry.
- We could also consider having `current_weather` return an array instead of string, but the string works well.

### DIFF
--- a/examples/assistant_chat.rb
+++ b/examples/assistant_chat.rb
@@ -1,23 +1,9 @@
 require "dotenv/load"
 require "langchain"
 require "openai"
-require "reline"
-
-# gem install reline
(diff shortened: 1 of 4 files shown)


andreibondarev:
@mattlindsey Thank you for your work here. I went ahead and rewrote the whole tool: https://github.com/patterns-ai-core/langchainrb/pull/728.

PR #630, opened 2024-05-20
aws bedrock streaming

agomezcampero:
This PR adds streaming to Bedrock API similar to how it's handled in OpenAI. If no block is passed, the code works exactly the way it used to.

Usage:
```
Langchain::LLM::AwsBedrock.new.chat(messages:, model:) do |chunk|
  do_something(chunk)
end
```

image


lib/langchain/llm/aws_bedrock.rb
```diff
@@ -260,5 +281,29 @@ def compose_parameters_ai21(params)
         }
       }
     end
+
+    def response_from_chunks(chunks)
```
andreibondarev, line 285: Do you think you could add a unit test just this method?

agomezcampero: sure!

andreibondarev:
@agomezcampero Great job, thank you!

PR #631, opened 2024-05-20
Add embedding method for Langchain::LLM::GoogleGemini

swerner:


lib/langchain/llm/response/google_gemini_response.rb
```diff
@@ -27,7 +27,11 @@ def embedding
     end
 
     def embeddings
-      [raw_response.dig("predictions", 0, "embeddings", "values")]
+      if raw_response.key?("embedding")
```
swerner, line 30: Had to do something weird here since the GoogleGeminiResponse class was also used by GoogleVertexAI and the response of the embeddings is slightly different, see `spec/fixtures/llm/google_gemini/embed.json` below...

spec/fixtures/llm/google_gemini/embed.json
```diff
@@ -0,0 +1,9 @@
+{
+  "embedding": {
+    "values": [
+      0.013168523,
```
swerner, line 4: truncated the embedding values here because it looks like you also did that for the Vertex fixtures. Thinking that's ok?

andreibondarev:
@swerner `standardrb --fix` and let's ship it! Thank you for the PR!

PR #633, opened 2024-05-22
Remove check support assistant message

tianlu1677:
Because when we use other message type, but check support LLM, so someone can not implement other llm.

If someboy use not exist message, will raise error default. 

```
Langchain::Messages::XXMessage
```


andreibondarev:
@tianlu1677 Sorry -- could you please explain what this fixes?

andreibondarev:
Closing since the we don't have an LLM check anymore.

PR #635, opened 2024-05-22
Support Ruby 3.0 again

alvaromat:
Support using the gem with ruby 3.0, preventing users to stick with outdated gem versions due to their current Ruby version.

Also, I've removed dependency to active support 7, as It prevents you to add the gem to a Rails 6 project. 
Is ActiveSupport version 7 needed? Can't older versions be used?

Issue: https://github.com/patterns-ai-core/langchainrb/issues/634


andreibondarev:
@alvaromat I went ahead and removed activesupport in this [PR](https://github.com/patterns-ai-core/langchainrb/pull/639).

andreibondarev:
@alvaromat Could you please check why the spec is failing?

PR #644, opened 2024-05-29
Support Ollama streamed responses

### LINKED ISSUE #550: Fix `Ollama#chat()` streaming

dferrazm:
- `chat` : Added support for streamed responses.
- `complete` : Fixed. It was not working at all. Now it works with both streamed and non-streamed responses.

To generate non-streamed responses, call the methods without passing a block. To generate streamed responses, passes the block. Eg.

```rb
# non-streamed resps
resp = ollama.chat(messages: [{ role: "user", content: "Hi" }])
resp.chat_completion # => "Hello there!"

# streamed resps
resp = ollama.chat(messages: [{ role: "user", content: "Hi" }]) { |resp| print resp.chat_completion }
# Will print iteratively "Hello there!"
resp.chat_completion # => "Hello there!"
```

Note: Passing the `stream` paramater to method will not have any effect anymore.

Closes #550


dferrazm:
Missing tests.

lib/langchain/llm/ollama.rb
```diff
@@ -265,5 +261,40 @@ def client
...
     end
+
+    def json_responses_chunk_handler(&block)
+      proc do |chunk, _size|
+        chunk.split("\n").each do |chunk_line|
+          parsed_chunk = begin
+            JSON.parse(chunk_line)
+          rescue JSON::ParserError
+            # In some cases the chunk exceeds the buffer size and the JSON parser fails.
+            # TODO: How to better handle this?
```
dferrazm, line 272: it's not clear what we should do here @andreibondarev . wdyt?

lib/langchain/llm/aws_bedrock.rb
```diff
@@ -136,7 +136,7 @@ def complete(prompt:, **params)
     # @option params [Float] :top_p Use nucleus sampling.
     # @option params [Integer] :top_k Only sample from the top K options for each subsequent token
     # @yield [Hash] Provides chunks of the response as they are received
-    # @return [Langchain::LLM::AnthropicMessagesResponse] Response object
+    # @return [Langchain::LLM::AnthropicResponse] Response object
```
andreibondarev, line 139: Nice find

lib/langchain/llm/ollama.rb
```diff
@@ -132,29 +131,19 @@ def complete(
...
-              nil
-            end
-
-            response += json_chunk.dig("response") unless json_chunk.blank?
-          end
+        req.options.on_data = json_responses_chunk_handler do |parsed_chunk|
+          responses_stream << parsed_chunk
 
-          yield json_chunk, size if block
+          yield Langchain::LLM::OllamaResponse.new(parsed_chunk, model: parameters[:model]) if block_given?
```
andreibondarev, line 142: Maybe we start referencing these classes directly here since it's all under the same namespace: `OllamaResponse.new ...`. It's cleaner that way, right?

dferrazm: yep. it'll probably work too.

lib/langchain/llm/ollama.rb
andreibondarev: I wonder if the error just needs to be thrown? There's no graceful way to recover from this and still provide a useful response, I don't think?

lib/langchain/llm/response/ollama_response.rb
```diff
@@ -38,15 +36,19 @@ def role
...
-      raw_response.dig("eval_count")
+      raw_response.dig("eval_count") if done?
     end
 
     def total_tokens
-      prompt_tokens + completion_tokens
+      prompt_tokens + completion_tokens if done?
+    end
+
+    def done?
```
andreibondarev, line 50: `private`?

andreibondarev:
@dferrazm I just tried this but nothing was streamed to me:
```ruby
irb(main):005> llm = Langchain::LLM::Ollama.new(url: ENV["OLLAMA_URL"])
=>
#<Langchain::LLM::Ollama:0x0000000128a3a3d0
...
irb(main):006* llm.chat messages: [{role:"user", content:"hey"}] do |chunk|
irb(main):007*   puts chunk
irb(main):008> end
#<Langchain::LLM::OllamaResponse:0x000000012803b988>
=>
#<Langchain::LLM::OllamaResponse:0x0000000128038170
 @model="llama3",
 @prompt_tokens=nil,
 @raw_response=
  {"model"=>"llama3",
   "created_at"=>"2024-05-29T15:56:04.473077Z",
   "message"=>{"role"=>"assistant", "content"=>"Hey! How's it going?"},
   "done_reason"=>"stop",
   "done"=>true,
   "total_duration"=>11215288666,
   "load_duration"=>10900135000,
   "prompt_eval_count"=>11,
   "prompt_eval_duration"=>152727000,
   "eval_count"=>8,
   "eval_duration"=>156123000}>
   ```

dferrazm:
@andreibondarev you have to pass `stream: true`. Should it be set by default? Probably fallback to `true` if the block is given, right?

lib/langchain/llm/ollama.rb
dferrazm: updated

lib/langchain/llm/response/ollama_response.rb
dferrazm: updated

lib/langchain/llm/ollama.rb
dferrazm: yeah, let's remove that and let it blow up, until we figure out a way to handle this.

PR #645, opened 2024-05-29
OpenAI response with tools

agomezcampero:
Currently if you call `llm.chat` with tools and streaming the tool_calls don't show up in the raw_response. I'm adding the `tool_calls` to the `response_from_chunks` so the final response when using streaming is almost the same as when not using streaming.
image


andreibondarev (approved):
@agomezcampero Looks great, thank you!

PR #646, opened 2024-05-30
Wrap OpenAI streaming responses

andreibondarev:


sergiobayona:
@andreibondarev PR seems stalled. does it need more work? or close?

PR #649, opened 2024-05-31
Exclude tiktoken_ruby from automatic max_tokens validation

andreibondarev:


night91:
@andreibondarev This is crashing: `cannot load such file -- tiktoken_ruby` as https://github.com/patterns-ai-core/langchainrb/blob/main/lib/langchain/utils/token_length/openai_validator.rb has a require for require "tiktoken_ruby".

andreibondarev:
@night91 I'll take a look at it later today. If you need it sooner -- feel free to open a PR with a fix.

PR #653, opened 2024-06-03
fix undefined text completions in anthropic response

agomezcampero:
Really small bug fix. Anthropic can return only tool_calls, if that is the case then you don't have any text completions and you get an error trying to get the `chat_completion`


andreibondarev:
@agomezcampero Thank you for the fix!

andreibondarev (approved):

PR #660, opened 2024-06-06
accept non ascii url

michelson:
for some url like `"https://vadb.org/events/d/México"` the url will not work, with this change , it will

### DIFF
--- a/lib/langchain/loader.rb
+++ b/lib/langchain/loader.rb
@@ -90,7 +90,7 @@ module Langchain
     private
 
     def load_from_url
-      URI.parse(@path).open
+      URI.parse(URI::DEFAULT_PARSER.escape(@path)).open
     end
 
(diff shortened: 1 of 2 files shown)


lib/langchain/loader.rb
```diff
@@ -90,7 +90,7 @@ def load(&block)
     private
 
     def load_from_url
-      URI.parse(@path).open
+      URI.parse(URI::DEFAULT_PARSER.escape(@path)).open
```
andreibondarev, line 93: @michelson Think we could add a quick spec for this?

michelson: spec added, not sure if that tests the whole thing,but I have the fork in production and works

andreibondarev:
@michelson Thank you for the PR!

PR #665, opened 2024-06-10
fix: GoogleGemini generation_config param

### LINKED ISSUE #648: Google Gemini LLM generation config problem
The following line in Google Gemini LLM seems to be wrong.
If the `temperature` is defined, the whole `generation_config` object gets replaced thus ignoring any pre defined parameters there!

https://github.com/patterns-ai-core/langchainrb/blob/0efc39269e5c2c123886fa34e10b28c682ada948/lib/langchain/llm/google_gemini.rb#L45

mazenkhalil:
Fixes #648

### DIFF
--- a/lib/langchain/llm/google_gemini.rb
+++ b/lib/langchain/llm/google_gemini.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require "active_support/core_ext/hash/keys"
+require "active_support/core_ext/string/inflections"
+
 module Langchain::LLM
   # Usage:
   #     llm = Langchain::LLM::GoogleGemini.new(api_key: ENV['GOOGLE_GEMINI_API_KEY'])
@@ -18,7 +21,9 @@ module Langchain::LLM
 
       chat_parameters.update(
         model: {default: @defaults[:chat_completion_model_name]},
-        temperature: {default: @defaults[:temperature]}
+        temperature: {default: @defaults[:temperature]},
+        generation_config: {default: nil},
+        safety_settings: {default: nil}
       )
       chat_parameters.remap(
         messages: :contents,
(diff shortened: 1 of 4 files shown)


andreibondarev:
@mazenkhalil Do you think you could add a quick spec?

mazenkhalil:
> @mazenkhalil Do you think you could add a quick spec?

I have synced the whole implementation with GeminiAPI & and I have added the tests

lib/langchain/llm/google_gemini.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require "active_support/core_ext/hash/keys"
+require "active_support/core_ext/string/inflections"
```
andreibondarev, line 4: Do we need to use `active_support`? I specifically [removed it](https://github.com/patterns-ai-core/langchainrb/pull/639) to reduce the number of dependencies.

mazenkhalil: Two utilities have been used. I am not much familiar with Ruby. if there's another way to implement that, I'm open to suggestions

lib/langchain/llm/google_gemini.rb
```diff
@@ -42,13 +47,25 @@ def chat(params = {})
...
+      parameters.delete(:response_format)
+      parameters[:generation_config][:stop_sequences] ||= parameters[:stop] if parameters[:stop]
+      parameters.delete(:stop)
 
       uri = URI("https://generativelanguage.googleapis.com/v1beta/models/#{parameters[:model]}:generateContent?key=#{api_key}")
 
       request = Net::HTTP::Post.new(uri)
       request.content_type = "application/json"
-      request.body = parameters.to_json
+      request.body = parameters.deep_transform_keys { |key| key.to_s.camelize(:lower).to_sym }.to_json
```
andreibondarev, line 68: Do we actually need to transform the keys here? Does it not work otherwise?

lib/langchain/llm/google_gemini.rb
mazenkhalil: I've created internal utils class

lib/langchain/llm/google_gemini.rb
mazenkhalil: It's important to note that merely having a working API implementation doesn't guarantee its stability over time. The best approach is to strictly adhere to the API specifications. Failure to do so may result in the API discontinuing support unexpectedly.

mazenkhalil:
@andreibondarev Is there anything I can do to speed up the review process here?

lib/langchain/llm/google_gemini.rb
```diff
@@ -42,13 +44,25 @@ def chat(params = {})
       raise ArgumentError.new("messages argument is required") if Array(params[:messages]).empty?
 
       parameters = chat_parameters.to_params(params)
-      parameters[:generation_config] = {temperature: parameters.delete(:temperature)} if parameters[:temperature]
+      parameters[:generation_config] ||= {}
+      parameters[:generation_config][:temperature] ||= parameters[:temperature] if parameters[:temperature]
+      parameters.delete(:temperature)
```
andreibondarev, line 49: Why not change:
```ruby
parameters[:generation_config][:temperature] ||= parameters[:temperature] if parameters[:temperature]
parameters.delete(:temperature)
```
to
```ruby
parameters[:generation_config][:temperature] ||= parameters.delete(:temperature) if parameters[:temperature]
```
everywhere?

lib/langchain/utils/hash_transformer.rb
```diff
@@ -0,0 +1,25 @@
+module Langchain
+  module Utils
+    class HashTransformer
```
andreibondarev, line 3: I'm wondering if we should've just yanked out the code out of the `activesupport` lib? This is fine, but I was wondering if it could be more elegant.

andreibondarev:
> @andreibondarev Is there anything I can do to speed up the review process here?

A few comments, otherwise -- let's ship it!

lib/langchain/llm/google_gemini.rb
mazenkhalil: Because we always need to delete the parameter even if it is already defined in the generation_config. Otherwise actual payload will have additional unneeded parameters. This case already covered in the specs.

lib/langchain/utils/hash_transformer.rb
mazenkhalil: The code here is 100% genuine, not copied from ActiveSupport in any way. In fact It is GPT generated.

andreibondarev:
@mazenkhalil Thank you! Merged.

Issue #566, opened 2024-04-06
Add a lean state machine for Assistant's phases.


andreibondarev:
## Description

The state machine should support the different states that the Assistant goes through. The [OpenAI Assistants one](https://platform.openai.com/docs/assistants/how-it-works/runs-and-run-steps) is a good inspiration and first step but need to be cautious to make it generic enough to account for other LLMs and workflows.

dghirardo:
Hi @andreibondarev! I'm currently working on this task. You can assign it to me.

andreibondarev:
> Hi @andreibondarev! I'm currently working on this task. You can assign it to me.

I'm curious whether you think we should be using a library here, and which one.

andreibondarev:
Complete.

PR #677, opened 2024-06-24
State machine for Assistant's execution phases

dghirardo:
### Issue #566

This pull request introduces a simplified state machine for managing the Assistant's execution phases. Inspired by the [OpenAI Assistants](https://platform.openai.com/docs/assistants/how-it-works/runs-and-run-steps) model, this leaner approach enhances code readability and simplifies debugging.

**Changes**:
- [x] Introduced state management within the run method with possible states: `in_progress`, `running_tools`, `completed`, `failed`, and `expired`.
- [x] Refactored code into distinct methods following the Single Responsibility Principle (SRP), enhancing readability and ease of debugging.
- [x] Used `case` statements for state management to improve code clarity and avoid nested `if` statements.
- [x] Created a `standard_role` method for consistent message type handling across different LLMs, enabling `case` statements in `process_latest_message` to manage the execution flow based on the role of the last message.
- [x] Incorporated an `expired` status similar to OpenAI, transitioning to expired if tool output is not provided within 10 minutes to prevent application hang.
- [x] Fixed an issue where the content of a message was not included if the LLM responded with one or more tool calls.

**About implementation**:

After analyzing OpenAI's approach, I believe a specialized state machine gem is over-engineered for our needs. OpenAI initializes a thread and creates a job for each run, allowing state access during execution, which is ideal for asynchronous processing. In our synchronous context, a simpler state machine within the run method is sufficient, reducing complexity while maintaining clear state management.

**Notes**:

I have left several TODO notes in the code for further improvements and would appreciate your feedback on these points. I am open to discussing the implementation and any potential improvements you might suggest. Thank you!

### DIFF
--- a/lib/langchain/assistants/assistant.rb
+++ b/lib/langchain/assistants/assistant.rb
@@ -79,54 +79,10 @@ module Langchain
         return
       end
 
-      running = true
+      state = :in_progress
+      state = handle_state(state, auto_tool_execution) until run_finished?(state)
 
-      while running
-        # TODO: I think we need to look at all messages and not just the last one.
-        last_message = thread.messages.last
-
-        if last_message.system?
-          # Do nothing
-          running = false
-        elsif last_message.llm?
-          if last_message.tool_calls.any?
-            if auto_tool_execution
-              run_tools(last_message.tool_calls)
-            else
-              # Maybe log and tell the user that there's outstanding tool calls?
-              running = false
-            end
-          else
-            # Last message was from the assistant without any tools calls.
-            # Do nothing
-            running = false
-          end
-        elsif last_message.user?
-          # Run it!
-          response = chat_with_llm
-
-          if response.tool_calls.any?
-            # Re-run the while(running) loop to process the tool calls
-            running = true
-            add_message(role: response.role, tool_calls: response.tool_calls)
-          elsif response.chat_completion
-            # Stop the while(running) loop and add the assistant's response to the thread
-            running = false
-            add_message(role: response.role, content: response.chat_completion)
-          end
-        elsif last_message.tool?
-          # Run it!
-          response = chat_with_llm
-          running = true
-
-          if response.tool_calls.any?
-            add_message(role: response.role, tool_calls: response.tool_calls)
-          elsif response.chat_completion
-            add_message(role: response.role, content: response.chat_completion)
-          end
-        end
-      end
-
+      # TODO: Should we return the final state along with the messages?
       thread.messages
     end
 
@@ -146,13 +102,7 @@ module Langchain
     # @param output [String] The output of the tool
     # @return [Array<Langchain::Message>] The messages in the thread
     def submit_tool_output(tool_call_id:, output:)
-      tool_role = if llm.is_a?(Langchain::LLM::OpenAI)
-        Langchain::Messages::OpenAIMessage::TOOL_ROLE
-      elsif [Langchain::LLM::GoogleGemini, Langchain::LLM::GoogleVertexAI].include?(llm.class)
-        Langchain::Messages::GoogleGeminiMessage::TOOL_ROLE
-      elsif llm.is_a?(Langchain::LLM::Anthropic)
-        Langchain::Messages::AnthropicMessage::TOOL_ROLE
-      end
+      tool_role = determine_tool_role
 
       # TODO: Validate that `tool_call_id` is valid by scanning messages and checking if this tool call ID was invoked
       add_message(role: tool_role, content: output, tool_call_id: tool_call_id)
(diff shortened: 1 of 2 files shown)


lib/langchain/assistants/assistant.rb
```diff
@@ -79,54 +79,10 @@ def run(auto_tool_execution: false)
...
-            add_message(role: response.role, tool_calls: response.tool_calls)
-          elsif response.chat_completion
-            add_message(role: response.role, content: response.chat_completion)
-          end
-        end
-      end
+      state = :in_progress
+      state = handle_state(state, auto_tool_execution) until run_finished?(state)
 
+      # TODO: Should we return the final state along with the messages?
```
andreibondarev, line 85: I think it's okay to skip. But I think the developer should be able to access it with:
```ruby
assistant.state
#=> :in_progress
```
What do you think?

lib/langchain/assistants/assistant.rb
```diff
@@ -183,6 +133,128 @@ def instructions=(new_instructions)
...
+    # Handle the current state and transition to the next state
+    #
+    # @param state [Symbol] The current state
+    # @param auto_tool_execution [Boolean] Whether or not to automatically run tools
+    # @return [Symbol] The next state
+    def handle_state(state, auto_tool_execution)
+      case state
+      when :in_progress
+        process_latest_message(auto_tool_execution)
+      when :running_tools
```
andreibondarev, line 153: I'm wondering if we should just copy how OpenAI calls it and name it `:requires_action` to be more aligned with that interface?

lib/langchain/assistants/assistant.rb
```diff
@@ -183,6 +133,128 @@ def instructions=(new_instructions)
...
+        execute_tools
+      else
+        Langchain.logger.error("Unexpected state encountered: #{state}")
```
andreibondarev, line 156: What are some examples of when the execution would go here?

lib/langchain/assistants/assistant.rb
```diff
@@ -183,6 +133,128 @@ def instructions=(new_instructions)
...
+        :failed
+      end
+    end
+
+    # Execute the tools based on the tool calls in the last message
+    #
+    # @return [Symbol] The next state
+    def execute_tools
+      # TODO: Should we create a method parameter to let the user change the value of the tool timeout?
+      Timeout.timeout(600) { run_tools(thread.messages.last.tool_calls) }
```
andreibondarev, line 233: This is 600 seconds so 10 minutes, right? I was thinking of starting very lean and not putting these kinds of guardrails just yet.

andreibondarev:
@dghirardo First off -- thank you for this PR! 🎉❤️ 

I left you a few comments but I had pulled it out, tested it out, and it was looking pretty good!

lib/langchain/assistants/assistant.rb
dghirardo: Yes, adding a `state` attribute to the assistant could be a valid solution without modifying the return value of the method.

A note: In synchronous execution, `assistant.state` will always return the final state (`:completed`, `:failed`, or `:expired`) because during the `run` method, you cannot access the attribute. Therefore, you will never see `:in_progress` or `:running_tools`. Correct?

lib/langchain/assistants/assistant.rb
dghirardo: Yes, sure, I agree!

lib/langchain/assistants/assistant.rb
dghirardo: Yes, it's 10 minutes. I took inspiration from OpenAI's assistant, which does the same. If you confirm, I can remove this check and the `:expired` state as well.

lib/langchain/assistants/assistant.rb
andreibondarev: Wouldn't you see the `:running_tools` (`:requires_action`)? For example:
```ruby
irb(main):019> a.add_message content:"Latests news in politics?"
=>
[#<Langchain::Messages::OpenAIMessage:0x000000011f1de708 @content="You are a news reporter", @role="system", @tool_call_id=nil, @tool_calls=[]>,
 #<Langchain::Messages::OpenAIMessage:0x00000001202b2ae8 @content="Latests news in politics?", @role="user", @tool_call_id=nil, @tool_calls=[]>]
irb(main):020> a.run auto_tool_execution: false
I, [2024-06-25T13:10:51.775951 #77146]  INFO -- : [Langchain.rb] [Langchain::Assistant]: Sending a call to Langchain::LLM::OpenAI
=>
[#<Langchain::Messages::OpenAIMessage:0x000000011f1de708 @content="You are a news reporter", @role="system", @tool_call_id=nil, @tool_calls=[]>,
 #<Langchain::Messages::OpenAIMessage:0x00000001202b2ae8 @content="Latests news in politics?", @role="user", @tool_call_id=nil, @tool_calls=[]>,
 #<Langchain::Messages::OpenAIMessage:0x000000012011a370
  @content="",
  @role="assistant",
  @tool_call_id=nil,
  @tool_calls=
   [{"id"=>"call_MnJgY7DHwfPT1yPXRM1KRF1H",
     "type"=>"function",
     "function"=>{"name"=>"news_retriever__get_top_headlines", "arguments"=>"{\"category\":\"general\",\"q\":\"politics\",\"page_size\":5}"}}]>]
irb(main):021>
```
At this point the `state` would be `:running_tools` (or `:requires_action`), correct?

andreibondarev:
@dghirardo I think I just realized our slight misunderstanding here. The way that I was thinking about this is:
```
if (LLM returns tool_calls)
  state = :requires_action
  if auto_tool_execution == true
    # We automatically execute the tools
  else
    # Return the control back to the developer
```

The developer has an option to execute the tools manually and submit the output, e.g.: `assistant.submit_tool_output(tool_call_id:, output:)`

lib/langchain/assistants/assistant.rb
dghirardo: No, as you can see in `handle_llm_message`, if auto tool execution is disabled, the return value is currently `:completed`. This is necessary to exit the loop (see `run_finished?`).

```ruby
if last_message.tool_calls.any?
  auto_tool_execution ? :running_tools : :completed
else
```

As written in the TODO notes, I was thinking to implement a new state (like `:requires_manual_action`). Or we should find a way to exit while keeping the state `requires_action`. What do you think?

andreibondarev: @dghirardo My gut feeling is to simplify, so less states. Do you agree? It's much easier to add new states than to try to simplify it later.

lib/langchain/assistants/assistant.rb
dghirardo: I removed it. There can't be any other state different from the ones we have decided.

lib/langchain/assistants/assistant.rb
dghirardo: One last question. When the assistant is initialized for the first time, what will `state` be set to? Nil?

andreibondarev: Good question. Either "new" or "initialized" or "ready". What do you think?

dghirardo:
Hi @andreibondarev, I just made a commit with the latest changes we discussed in the past few days. We should be good now!

andreibondarev:
@dghirardo Wow, amazing work!

PR #695, opened 2024-07-04
Instructions are ignored for Anthropic if tools are not used

### LINKED ISSUE #694: Anthropic instructions are ignored unless using tools
**Describe the bug**
Anthropic instructions are not added to the chat parameters unless tools are used.

**To Reproduce**
Use Anthropic assistant with instructions but without tools. Chat will not receive the instructions.

**Expected behavior**
Instructions should be used regardless of tool presence.

victorivanov:
Fixes #694

Note: based on the code, the same issue might be affecting Gemini and Vertex AI but I'm not familiar with those.

### DIFF
--- a/lib/langchain/assistants/assistant.rb
+++ b/lib/langchain/assistants/assistant.rb
@@ -253,21 +253,25 @@ module Langchain
 
       params = {messages: thread.array_of_message_hashes}
 
-      if tools.any?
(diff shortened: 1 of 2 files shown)


andreibondarev:
@victorivanov Looks great! I had tested it with Google Gemini.

PR #703, opened 2024-07-09
Revert Style/ArgumentsForwarding Rubocop rule

andreibondarev:


.rubocop.yml
```diff
@@ -1,6 +1,9 @@
 AllCops:
   SuggestExtensions: false
 
+Style/ArgumentsForwarding:
```
Rukomoynikov, line 4: I can be wrong, but could you please try to move `Style/ArgumentsForwarding` below all other instructions in this file?

Rukomoynikov: As they can overwrite the instructions above, including `Style/ArgumentsForwarding`

Rukomoynikov: It didn't help. I tried. :)

andreibondarev: @Rukomoynikov It's very odd. I don't quite understand why these rules are being ignored in `.rubocop.yml` or `.standard.yml`.

andreibondarev:
@Rukomoynikov Looks like you fixed it! 👀 🎉

Rukomoynikov:
> @Rukomoynikov Looks like you fixed it! 👀 🎉

Thank you, but I just slightly altered your solution.

PR #659, opened 2024-06-05
Improve LLM model tool annotation

dghirardo:
### Issue #637

This pull request implements the `tool_tailor` gem (https://github.com/kieranklaassen/tool_tailor) to automate the generation of tool annotations.

**Changes**:

- [x] All previously used JSON files for tool annotations have been removed.
- [x] Tool-specific directories have been eliminated, resulting in a cleaner and more straightforward directory structure.
- [x] The `:ANNOTATIONS_PATH` constant has been replaced with `:FUNCTIONS`, an array of symbols corresponding to the tool's functions.
- [x] Methods for validating constants (`:NAME`, `:FUNCTIONS`) have been implemented to ensure consistency and correctness.

Issue #637, opened 2024-05-22
WIP: Improve LLM model tool annotation 


andreibondarev:
WIP: Overload the `Module` class

Evaluate using https://github.com/kieranklaassen/tool_tailor

dghirardo:
Hi @andreibondarev, you can assign this task to me. I'm already working on it. I will send the pull request by tomorrow

andreibondarev:
Fixed already.

PR #708, opened 2024-07-15
Rewrite Assistant tools definition for easier configuration

dghirardo:
### Issue #637
This pull request introduces a new approach to defining Assistant tools, simplifying the process and enhancing the developer experience.

**Changes**:
- [x] Removed the need for JSON files to define tool schemas, reducing configuration complexity.
- [x] Introduced a new base module `Langchain::ToolDefinition` for tool classes, replacing the previous `Langchain::Tool:Base` class.
- [x] Implemented an expressive and simple syntax for defining tool actions (methods), as suggested by @andreibondarev  in PR #659.
- [x] Added validators to action definition methods, providing immediate feedback and guiding developers in creating correct schemas.
- [x] Added tests for the new `Langchain::ToolDefinition` methods.

**Key methods**:
- `define_action`: Defines a new action for the tool
- `property`: Defines properties for the action parameters
- `item`: Alias for `property`, used for array items

**New tool creation process**:
1. Extend your class with `Langchain::ToolDefinition`
2. Use `define_action` to define each action of the tool

**Important note**:

Each `define_action` call must match a method in the tool class. For example, `:sample_action` in `define_action` should match a `sample_action` method in the class. The same applies to parameter names.

**Example implementation**:
```ruby
class MyTool
  extend Langchain::ToolDefinition

  define_action :sample_action, description: "Demonstrates various property types and configurations" do
    property :string_prop, type: "string", description: "A simple string property"
    property :number_prop, type: "number", description: "A number property"
    property :integer_prop, type: "integer", description: "An integer property"
    property :boolean_prop, type: "boolean", description: "A boolean property"
    property :enum_prop, type: "string", description: "An enum property", enum: ["option1", "option2", "option3"]
    property :required_prop, type: "string", description: "A required property", required: true
    property :array_prop, type: "array", description: "An array property" do
      item type: "string", description: "Array item"
    end
    property :object_prop, type: "object", description: "An object property" do
      property :nested_string, type: "string", description: "Nested string property"
      property :nested_number, type: "number", description: "Nested number property"
    end
  end

  def sample_action(string_prop:, number_prop:, integer_prop:, boolean_prop:, enum_prop:, required_prop:, array_prop:, object_prop:)
    # Implementation of the sample_action method
  end
end
```


lib/langchain/tool_definition.rb
```diff
@@ -0,0 +1,209 @@
+# frozen_string_literal: true
+
+require "json"
+
+#
+# Extends a class to be used as a tool in the assistant.
+# A tool is a collection of actions (methods) used to perform specific tasks.
```
andreibondarev, line 7: I'm wondering if we should say that "A tool is a collection of functions"?

lib/langchain/tool_definition.rb
```diff
@@ -0,0 +1,209 @@
...
+#
+# == Usage
+#
+# 1. Extend your class with {Langchain::ToolDefinition}
+# 2. Use {#define_action} to define each action of the tool
```
andreibondarev, line 12: I'm wondering if we should call this method `define_function` or `define_tool_function`. What do you think?

lib/langchain/tool_definition.rb
```diff
@@ -0,0 +1,209 @@
...
+# @example Defining a tool with various property types and configurations
+#   define_action :sample_action, description: "Demonstrates various property types and configurations" do
+#     property :string_prop, type: "string", description: "A simple string property"
+#     property :number_prop, type: "number", description: "A number property"
+#     property :integer_prop, type: "integer", description: "An integer property"
+#     property :boolean_prop, type: "boolean", description: "A boolean property"
+#     property :enum_prop, type: "string", description: "An enum property", enum: ["option1", "option2", "option3"]
+#     property :required_prop, type: "string", description: "A required property", required: true
+#     property :array_prop, type: "array", description: "An array property" do
+#       item type: "string", description: "Array item"
```
andreibondarev, line 31: Would this support an array of objects? Something like:
```ruby
property :array_prop, type: "array" do
  item type: "object"
```

lib/langchain/tool_definition.rb
```diff
@@ -0,0 +1,209 @@
...
+#     property :object_prop, type: "object", description: "An object property" do
+#       property :nested_string, type: "string", description: "Nested string property"
+#       property :nested_number, type: "number", description: "Nested number property"
+#     end
+#   end
+#
+module Langchain::ToolDefinition
+  # Defines an action for the tool
+  #
+  # @param method_name [Symbol] Name of the method to define
```
andreibondarev, line 42: Symbol or String?

lib/langchain/tool/calculator.rb
```diff
@@ -0,0 +1,38 @@
...
+  #     gem "google_search_results", "~> 2.0.0"
+  #
+  # Usage:
+  #     calculator = Langchain::Tool::Calculator.new
+  #
+  class Calculator
+    extend Langchain::ToolDefinition
+    include Langchain::DependencyHelper
+
+    define_action :execute, description: "Evaluates a pure math expression or if equation contains non-math characters (e.g.: \"12F in Celsius\") then it uses the google search calculator to evaluate the expression" do
```
andreibondarev, line 18: This produces the follow function name:
```ruby
langchain_tool_calculator__execute
```
I'm wondering whether it should be simpler, just the class name without the namespace:
```ruby
calculator__execute
```

lib/langchain/tool_definition.rb
dghirardo: Sure! There are no limits to nesting. Here is an example:

```ruby
define_action :sample_action, description: "Lorem ipsum" do
  property :array_prop, type: "array" do
    item type: "object" do
      property :nested_string, type: "string"
    end
  end
end
```
The previous definition generates this scheme:

```json
{
  "type": "function",
  "function": {
    "name": "tool_class__sample_action",
    "description": "Lorem ipsum",
    "parameters": {
      "type": "object",
      "properties": {
        "array_prop": {
          "type": "array",
          "items": {
            "type": "object",
            "properties": {
              "nested_string": {
                "type": "string"
              }
            },
            "required": []
          }
        }
      },
      "required": []
    }
  }
}
```

lib/langchain/tool_definition.rb
dghirardo: I have chosen to use a symbol. It seems the correct choice to me. Even Ruby uses symbols when referencing methods. For example, see [Object#method](https://ruby-doc.org/3.3.4/Object.html#method-i-method).

lib/langchain/tool_definition.rb
```diff
@@ -0,0 +1,209 @@
...
+  # @return [ActionSchemas] The ActionSchemas instance
+  def action_schemas
+    @action_schemas ||= ActionSchemas.new(tool_name)
+  end
+
+  # Returns the snake_case version of the class name as the tool's name
+  #
+  # @return [String] The snake_case version of the class name
+  def tool_name
+    @tool_name ||= name.gsub(/([A-Z])/, '_\1').gsub(/^_/, "").gsub("::", "").downcase
```
andreibondarev, line 60: There could be a bug here. Example:
```ruby
class Langchain::Tool::API1Interface; end

irb(main):023> "Langchain::Tool::API1Interface".gsub(/([A-Z])/, '_\1').gsub(/^_/, "").gsub("::", "").downcase
=> "langchain_tool_a_p_i1_interface"
```

lib/langchain/tool/calculator.rb
dghirardo: Yes, we could do something like:

```ruby
define_action :execute, for: "calculator", description: "..." do
  # ...
end
```

The "__execute" part would be added automatically based on the method name.

However, I have some doubts about this approach. It feels a bit odd to define the tool name for each function. You'd have to repeat the same name multiple times, and if you make a typo, you lose consistency.

Since we're already inside the tool class, I would expect Langchain to handle this automatically.

We could add another method like:

```ruby
define_tool :calculator
```

This way, you only write the name once. But it's an extra step to remember.

I prefer Langchain to automatically define the name, but if we have to choose between these two options, I would lean towards the second one, which is similar to the old approach.

lib/langchain/tool_definition.rb
dghirardo: Thanks for pointing this out! It's not really a problem, the mechanism still works. But if we decide to go with the automatic naming approach, it's better to fix it for easier reading.

andreibondarev: The (possible) problem here is that the LLM uses this name as context as well, so if the name is malformed then it might mess up the LLM. Obviously the best function name is one that is descriptive to what the function actually does.

lib/langchain/tool/calculator.rb
andreibondarev: Maybe what we do is this:
```ruby
class Langchain::Tool::Calculator
  define_action :execute

  def execute()
```
becomes `name: "langchain_tool_calculator__execute"`

But you can also have an option to override the name completely with:
```ruby
class Langchain::Tool::Calculator
  define_action :execute, name: "calc__execute"

  def execute()
```
becomes `name: "calc__execute"`

What do you think?

lib/langchain/tool_definition.rb
dghirardo: Ok, I had missed this point! Thank you

dghirardo: Fixed inspired by the `underscore` method from `Active Support Inflector`

lib/langchain/tool/calculator.rb
dghirardo: After a discussion with @andreibondarev, it has been decided to keep the current functionality unchanged for now.

PR #712, opened 2024-07-25
Update default chat completions model to gpt-4o-mini

edudepetris:
This PR updates the default chat completions model to `gpt-4o-mini` as per [Open AI doc](https://platform.openai.com/docs/models/gpt-3-5-turbo)

### DIFF
--- a/spec/langchain/llm/openai_spec.rb
+++ b/spec/langchain/llm/openai_spec.rb
@@ -204,10 +204,10 @@ RSpec.describe Langchain::LLM::OpenAI do
   describe "#complete" do
     let(:response) do
       {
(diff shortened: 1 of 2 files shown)


spec/langchain/llm/openai_spec.rb
```diff
@@ -204,10 +204,10 @@
   describe "#complete" do
     let(:response) do
```
edudepetris, line 205: response taken from:

```rb
openai = Langchain::LLM::OpenAI.new(api_key: ENV["OPENAI_API_KEY"])
parameters = {
  n: 1,
  model: "gpt-4o-mini",
  messages: [{content: "Hello World", role: "user"}],
  temperature: 0.0
}
openai.chat(parameters)
```

PR #718, opened 2024-07-30
Send nonCamelCased tools params

Rukomoynikov:
Google Gemini or better say news_retriever tool didn't work due to the updates in the list of arguments. See the error message:

```
E, [2024-07-30T08:05:04.021082 #6] ERROR -- : [Langchain.rb] Error running tools: unknown keyword: :sortBy; /usr/src/app/lib/langchain/tool/news_retriever.rb:61:in
```

### DIFF
--- a/spec/langchain/llm/google_gemini_spec.rb
+++ b/spec/langchain/llm/google_gemini_spec.rb
@@ -39,7 +39,21 @@ RSpec.describe Langchain::LLM::GoogleGemini do
   describe "#chat" do
     let(:messages) { [{role: "user", parts: [{text: "How high is the sky?"}]}] }
     let(:raw_chat_completions_response) { double(body: File.read("spec/fixtures/llm/google_gemini/chat.json")) }
-    let(:params) { {messages: messages, model: "gemini-1.5-pro-latest", system: "system instruction", tool_choice: "AUTO", tools: [{name: "tool1"}], temperature: 1.1, response_format: "application/json", stop: ["A", "B"], generation_config: {temperature: 1.7, top_p: 1.3, response_schema: {"type" => "object", "description" => "sample schema"}}, safety_settings: [{category: "HARM_CATEGORY_UNSPECIFIED", threshold: "BLOCK_ONLY_HIGH"}]} }
+    let(:params) {
+      {
+        messages: messages,
+        model: "gemini-1.5-pro-latest",
+        system: "system instruction",
+        tool_choice: "AUTO",
+        tools: [{name: "news_retriever",
+                 parameters: {type: "object", properties: {sort_by: {type: "string"}}}}],
+        temperature: 1.1,
+        response_format: "application/json",
+        stop: ["A", "B"],
+        generation_config: {temperature: 1.7, top_p: 1.3, response_schema: {"type" => "object", "description" => "sample schema"}},
+        safety_settings: [{category: "HARM_CATEGORY_UNSPECIFIED", threshold: "BLOCK_ONLY_HIGH"}]
+      }
+    }
 
     before do
       allow(Net::HTTP).to receive(:start).and_return(raw_chat_completions_response)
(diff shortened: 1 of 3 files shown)


lib/langchain/tool/news_retriever.rb
```diff
@@ -20,9 +20,9 @@ class NewsRetriever
       property :from, type: "string", description: "A date and optional time for the oldest article allowed. This should be in ISO 8601 format"
       property :to, type: "string", description: "A date and optional time for the newest article allowed. This should be in ISO 8601 format"
       property :language, type: "string", description: "The 2-letter ISO-639-1 code of the language you want to get headlines for", enum: ["ar", "de", "en", "es", "fr", "he", "it", "nl", "no", "pt", "ru", "sv", "ud", "zh"]
-      property :sort_by, type: "string", description: "The order to sort the articles in", enum: ["relevancy", "popularity", "publishedAt"]
       property :page_size, type: "integer", description: "The number of results to return per page (request). 5 is the default, 100 is the maximum"
       property :page, type: "integer", description: "Use this to page through the results if the total results found is greater than the page size"
+      property :sortBy, type: "string", description: "The order to sort the articles in", enum: ["relevancy", "popularity", "publishedAt"]
```
andreibondarev, line 25: I guess I would ask this question: If we're telling the LLM that the parameter is called "sort_by" then why is it sending a function call with "sortBy"? I've not seen it happen with gpt-4 but I have seen this happen with Gemini. 

It's fascinating that you can tell it about it's issue and it'll resend a call with the corrected param:
image

andreibondarev: @Rukomoynikov The answer might just be "Gemini is hallucinating" but I hope that's not the answer.

Rukomoynikov: Hi @andreibondarev 

I think I found the issue, but didn't solve.

The sort_by parameters gets covernted into sortBy here in the function below. I think Google is not hallucinating, the langchain lib sends instructs Gemini to use sortBy parameter. And I believe it may happen with other `NewsRetriever` params aka search_in, exclude_domains, page_size but it's not noticeable now as Google doesn't use it.
https://github.com/patterns-ai-core/langchainrb/blob/ff699356068bd7d6bc768e5518f6b4fdcbdfc90f/lib/langchain/llm/google_gemini.rb#L65

Rukomoynikov: This PR fixes the issue, though it's not the most elegant solution.

Rukomoynikov: Sorry for bothering you @andreibondarev maybe I could improve the solution? Do you think the direction is wrong?

spec/langchain/llm/google_gemini_spec.rb
Rukomoynikov, line 1: I added a tool with a snake_case parameter and it fails without changes from the PR.

andreibondarev:
@Rukomoynikov I'll take a look at this a bit later today.

Rukomoynikov:
> @Rukomoynikov I'll take a look at this a bit later today.

Sure. Here is a simple way to reproduce and confirm the bug:

1. Switch to main branch
2. In order to check parameters sent to Google add `byebug\debugger` to https://github.com/patterns-ai-core/langchainrb/blob/89a2cbffec353a2f4eeafb92318e2c41e035b45b/lib/langchain/llm/google_gemini.rb#L101
3. Get into console with gem's code available (`bin/console`) and run this:
```ruby
llm = Langchain::LLM::GoogleGemini.new(api_key: 'api_key')
news_retriever = Langchain::Tool::NewsRetriever.new(api_key: 'api_key')

Langchain::Assistant.new(
  llm: llm,
  tools: [news_retriever],
  instructions: 'You are a news reporter.'
).add_message_and_run content:"What's are latest news on the 2024 elections?",auto_tool_execution: true
```
----
So, expected params sent to Google should contain `sort_by`. But since the method below recursively converts all hash keys to camelCase it sends `sortBy`.
https://github.com/patterns-ai-core/langchainrb/blob/89a2cbffec353a2f4eeafb92318e2c41e035b45b/lib/langchain/llm/google_gemini.rb#L65

andreibondarev:
@Rukomoynikov Yep, you're right, this is definitely a bug:
```ruby
[3] pry(#<Langchain::LLM::GoogleGemini>)> Langchain::Utils::HashTransformer.camelize_lower("sort_by").to_sym
=> :sortBy
```

andreibondarev:
@Rukomoynikov I think it was an oversight on my side to let this change go through: https://github.com/patterns-ai-core/langchainrb/pull/665#discussion_r1634024204

I think just reverting that one change would be the proper fix.

andreibondarev:
@Rukomoynikov Closing this PR as I fixed it here: https://github.com/patterns-ai-core/langchainrb/pull/723

PR #721, opened 2024-08-05
Add Support for Ollama WebUI Security

rience:
We're using Ollama through Open WebUI interface that provides security through API keys, generated for individual users. This PR adds support for passing api key to `Langchain::LLM::Ollama` constructor - if set, it'll add "Authorization" header.


andreibondarev:
@rience Thank you!

PR #727, opened 2024-08-16
drop Colorize due to GPL license

Azdaroth:
Solves https://github.com/patterns-ai-core/langchainrb/issues/726

Based on the output when running specs, it seems to do exactly the same job, but it would be great to have another pair of eyes on that.


andreibondarev (approved):
@Azdaroth Nicely done!

andreibondarev:
@Azdaroth Let me know if you have any other issues!

PR #729, opened 2024-08-17
remove to_bool gem

Azdaroth:
License finder reported one more issue besides `colorize` - that `to_bool` has _undefined_ license (which is odd as it clearly has MIT license, but apparently something is off with the gem's metadata). Seems like the best solution was to drop the gem and just implement a simple utility function:

- it's inline with the direction mentioned here about reducing number of dependencies: https://github.com/patterns-ai-core/langchainrb/issues/588
- solves license issues
- removes monkeypatch that _to_bool_ introduces for all objects


andreibondarev:
@Azdaroth Looks good!

PR #731, opened 2024-08-21
Add `Assistant#add_messages()` method and fix `Assistant#messages=` method

andreibondarev:


lib/langchain/assistants/assistant.rb
```diff
@@ -69,6 +69,24 @@ def add_message(content: nil, role: "user", tool_calls: [], tool_call_id: nil)
...
+    def messages=(messages)
+      clear_thread!
+      add_messages(messages: messages)
+    end
+
+    # Add multiple messages to the thread
+    #
+    # @param messages [Array<Hash>] The messages to add
+    # @return [Array<Langchain::Message>] The messages in the thread
+    def add_messages(messages:)
```
dferrazm, line 84: PR looks good to me. The only question would be if we need this as a kwarg? Maybe one positional arg is already clear enough in this case:

```rb
add_messages([{ content: "Hello" }])
```

PR #732, opened 2024-08-22
Add fallback tool_call_id to Ollama tool calls 

eliasfroehner:
### Description
Currently, the response for Ollama tool calls lack a `tool_call_id`. This is necessary to support output from multiple tools, as seen in the templates of llama3.1 and mistral-nemo. To address this, we should use the responded `function_name` as a fallback for `tool_call_id`.

For a permanent fix for Ollama, please refer to my GitHub issue for Ollama: [#6462](https://github.com/ollama/ollama/issues/6462). My implementation is prepared for future adjustments by Ollama.

### Changes made
- Added fallback for `tool_call_id` for Ollama tool calls
- Used the responded `function_name` as the `tool_call_id`

### Why this change is needed
This change is essential to ensure that outputs from multiple tools can be properly identified and managed.

### How to test
1. Make a tool call using a supported Ollama model (llama3.1, ...).
2. Verify that the `tool_call_id` is present and matches the `function_name`.
3. Ensure that multiple tool calls can be distinguished by their `tool_call_id`.

### Checklist
- [x] Tests pass

Thank you for your review!


eliasfroehner:
The best solution will be to use the "/v1/chat/completions" endpoint of Ollama instead of /api/chat in Langchain::LLM::Ollama for tool use as pointed out by frob in my issue.

andreibondarev:
> The best solution will be to use the /v1/chat/completions endpoint of Ollama instead of /api/chat in Langchain::LLM::Ollama for tool use as pointed out by frob in my issue.

Does the completions endpoint even support the `tools:` parameter? I can see that the chat endpoint does: https://github.com/ollama/ollama/blob/main/docs/api.md#generate-a-chat-completion

andreibondarev:
@eliasfroehner Thank you for flagging this.

I'm wondering if we should switch [Ollama#chat()](https://github.com/patterns-ai-core/langchainrb/blob/main/lib/langchain/llm/ollama.rb#L181) to use `/v1/chat/completions` (OpenAI-compatible endpoint)? Are there any cons to using it?

eliasfroehner:
@andreibondarev I don't see any reasons against it. We would likely need to make some changes in the Langchain::LLM::Ollama class, such as the method #generate_final_chat_completion_response.

I now tried the Langchain::LLM::OpenAI class instead of Langchain::LLM::Ollama with my assistant against my local Ollama installation, and everything worked fine.

lib/langchain/assistants/assistant.rb
```diff
@@ -366,6 +366,8 @@ def extract_tool_call_args(tool_call:)
             tool_call_id = tool_call.dig("id")
 
             function_name = tool_call.dig("function", "name")
+            tool_call_id ||= function_name
```
andreibondarev, line 388: Thoughts on changing it up a bit to:
```ruby
function_name = tool_call.dig("function", "name")
tool_call_id = tool_call.dig("id") || function_name
```

eliasfroehner: That is a good improvement.

andreibondarev:
Closing due to inactivity. Feel free to reopen.

PR #743, opened 2024-08-27
Improve Database tool

andreibondarev:


lib/langchain/tool/database.rb
```diff
@@ -43,39 +47,66 @@ def initialize(connection_string:, tables: [], exclude_tables: [])
...
+    #
+    # @param input [String] SQL query to be executed
+    # @return [Array] Results from the SQL query
+    def execute(input:)
+      Langchain.logger.info("Executing \"#{input}\"", for: self.class)
+
+      db[input].to_a
+    rescue Sequel::DatabaseError => e
+      Langchain.logger.error(e.message, for: self.class)
+      e.message # Return error to LLM
```
andreibondarev, line 94: This is crucial because we return back the error to the LLM to attempt to self-correct.

PR #759, opened 2024-09-03
Fix annonymous block issue

gbs0:


andreibondarev:
@gbs0 Thank you! These linter errors are annoying.

PR #766, opened 2024-09-07
Allow setting response_object: {} parameter when initializing supported Langchain::LLM::* classes

andreibondarev:
Setting `default_options: {response_format: ... }` when supported classes are initialized:
```ruby
# Setting 
irb(main):002> llm = Langchain::LLM::MistralAI.new(api_key: ENV["MISTRAL_AI_API_KEY"], default_options: {response_format: {type: "json_object"}})
=>
#<Langchain::LLM::MistralAI:0x0000000125076888
```
# Passes that option to the chat calls. JSON format is returned:
```ruby
irb(main):003> llm.chat messages:[{role:"user", content:"Hello json!"}]
=>
#<Langchain::LLM::MistralAIResponse:0x000000012431be68
 @model=nil,
 @raw_response=
  {"id"=>"4837d4b042874aab8b5c29a1cd0bef99",
   "choices"=>[{"index"=>0, "message"=>{"role"=>"assistant", "content"=>"{\"message\": \"Hello! How can I assist you today?\"}", "tool_calls"=>nil}, "finish_reason"=>"stop", "logprobs"=>nil}]
   ```

### CODE REVIEW

File: lib/langchain/llm/openai.rb [hunk 1]
```diff
@@ -44,7 +44,8 @@ def initialize(api_key:, llm_options: {}, default_options: {})
         top_logprobs: {},
         n: {default: @defaults[:n]},
         temperature: {default: @defaults[:temperature]},
-        user: {}
+        user: {},
+        response_format: {default: @defaults[:response_format]}
```
codenamev, line 48: FYI: if you add a default, it will use it (even if it is `nil`).  If you made this just an empty hash it should work to solve #766 
```suggestion
        response_format: {}
```

codenamev: The reason for this is that the unified parameters will drop any argument without a value _unless_ it has a default value (which it will use).

bibstha:
Looking forward to this change, as looks like OpenAI's structured response is quite reliable.

andreibondarev:
> Looking forward to this change, as looks like OpenAI's structured response is quite reliable.

I'll try to get this out later tonight!

lib/langchain/llm/openai.rb
andreibondarev: @codenamev The problem is that initializing the client with the default_options does not then set those parameters on every consecutive call:
```ruby
irb(main):002> llm = Langchain::LLM::OpenAI.new(api_key: ENV["OPENAI_API_KEY"], default_options: { response_format: { type: "json_object" } })
=>
#<Langchain::LLM::OpenAI:0x0000000125ebc690
```

The `response_format: { type: "json_object" }` parameter is not automatically set in the following call:
```ruby
irb(main):003> llm.chat messages:[{role:"user", content:"hey json!"}]
=>
#<Langchain::LLM::OpenAIResponse:0x0000000125b7c3e0
 @model=nil,
 @raw_response=
  {"id"=>"chatcmpl-ABAha7eKlGmTw9QRSoPQPCAvoqe5E",
   ...
   "choices"=>[{"index"=>0, "message"=>{"role"=>"assistant", "content"=>"Hello! How can I assist you today?", "refusal"=>nil}, "logprobs"=>nil, "finish_reason"=>"stop"}],
```

codenamev: I see. The initializer should be able to mutate the library defaults.  I'll audit the LLM subclasses today.

PR #772, opened 2024-09-11
Update google_vertex_ai.rb

palladius:
Remove the tool_choice line as it breaks Gemini Function Calling. Removing it solves it.

### DIFF
--- a/lib/langchain/llm/google_vertex_ai.rb
+++ b/lib/langchain/llm/google_vertex_ai.rb
@@ -89,7 +89,7 @@ module Langchain::LLM
       params[:system] = {parts: [{text: params[:system]}]} if params[:system]
       params[:tools] = {function_declarations: params[:tools]} if params[:tools]
       # This throws an error when tool_choice is passed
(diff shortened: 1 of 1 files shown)


andreibondarev:
@palladius This is already fixed. Does Vertex AI API actually accept the tool_choice option?

palladius:
I don't know what to say. 

By googling it seems its an openAI thing: https://platform.openai.com/docs/guides/function-calling 

But it also seems to be supported by gemini, look: https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/function-calling 

so my tiepid answer wqould be YES.

palladius:
Note the code error is NOT at API level, it is at Ruby code level (trying to upcase a hash). 
So it seems to me a bug in your code, rather than a missing thing on Gemini side.

Can you explain to me what you were trying to achieve with  

```ruby
params[:tool_choice] = {function_calling_config: {mode: params[:tool_choice].upcase}} if params[:tool_choice]
``` 

? It seems to me like the  params[:tool_choice] you're actually passing is a hash instead of a string. Want me to file a bug to make it clearer?

andreibondarev:
This was already fixed.

PR #778, opened 2024-09-15
Update azure_spec.rb

scottwater:
Remove focus: true on Azure spec.


andreibondarev:
@scottwater Wow! What an omission on my end! Nice find!

PR #779, opened 2024-09-15
Add chat parameter support for steam_options

scottwater:
stream_options are use by OpenAI (others?) to include token usage when streaming chat responses.


lib/langchain/llm/parameters/chat.rb
```diff
@@ -21,6 +21,7 @@ class Chat < SimpleDelegator
 
       stop: {}, # multiple types (e.g. OpenAI also allows Array, null)
       stream: {}, # Enable streaming
+      stream_options: {}, # Enable usage data for streaming (at least with OpenAI)
```
andreibondarev, line 24: All of the parameters in this class are meant to be "common" parameters. Do any other LLMs accept this parameter or similar?

sergiobayona:
closing due to inactivity.

PR #796, opened 2024-09-27
Improved logging

dferrazm:
- Allow configuring the logger destination
- Use Langchain logger on OpenAI client requests
- Use Langchain logger on Google LLMs http requests
- Less verbose logger by replacing info msgs with debug


lib/langchain.rb
```diff
@@ -92,24 +91,54 @@
...
-
   module Errors
     class BaseError < StandardError; end
   end
+
+  module Colorizer
+    class << self
+      def red(str)
+        "\e[31m#{str}\e[0m"
+      end
```
andreibondarev, line 108: Consider maybe converting these methods to the following:
```ruby
COLORS = {
  red: 31,
  green: 32,
  yellow: 33,
  blue: 34
}

def wrap(color, str)
  "\e[#{COLORS[color]}m#{str}\e[0m"
end
```

lib/langchain/vectorsearch/epsilla.rb
```diff
@@ -127,7 +127,7 @@ def similarity_search_by_vector(embedding:, k: 4)
     # @param k [Integer] The number of results to have in context
     # @yield [String] Stream responses back one String at a time
     # @return [String] The answer to the question
-    def ask(question:, k: 4, &block)
+    def ask(question:, k: 4, &)
```
andreibondarev, line 130: I don't exactly understand why but under different versions of Ruby this keeps throwing an error when it's an "anonymous block"

lib/langchain.rb
andreibondarev: But this is your call!

andreibondarev:
@dferrazm Could you please add an entry to the `CHANGELOG.md`?

lib/langchain/vectorsearch/epsilla.rb
dferrazm: yeah. it's the auto linter. I'll revert it

andreibondarev (approved):

lib/langchain.rb
dferrazm: I tried out your suggestion, and I had to change other parts too. In the end the code was a bit less legible and clear. So I'm leaving as it is.

andreibondarev:
@dferrazm Great job! 👏🏻

PR #801, opened 2024-10-01
Resolve the issue of the AWS Bedrock chat model always returning `nil`

khaile:
The `Langchain::LLM::AwsBedrock#chat` cannot retrieve the model ID because of an incorrect key `chat_completion_model_name`.


lib/langchain/llm/aws_bedrock.rb
```diff
@@ -61,7 +61,7 @@ def initialize(completion_model: DEFAULTS[:completion_model_name], embedding_mod
         .merge(embedding_model_name: embedding_model)
 
       chat_parameters.update(
-        model: {default: @defaults[:chat_completion_model_name]},
+        model: {default: @defaults[:completion_model_name]},
```
andreibondarev, line 64: I think we actually need to add `chat_completion_model_name` key and value to the hash above:
```ruby
    DEFAULTS = {
      completion_model_name: "anthropic.claude-v2",
      embedding_model_name: "amazon.titan-embed-text-v1",
```

khaile: Looks like this 🤔 
CleanShot 2024-10-02 at 10 02 02

khaile:
@andreibondarev Please help to rereview 🙏🏽

lib/langchain/llm/aws_bedrock.rb
khaile: BTW, we are using the same `completion_provider` to compose parameters & response for both `chat` & `complete`. So I think we can use the same `completion_model_name` for both as well. WDYT?

andreibondarev:
@khaile Thank you for the PR!

PR #807, opened 2024-10-03
Move adapter-related classes to separate files

khaile:
- Create new adapters for various LLMs (Anthropic, Google Gemini, MistralAI, Ollama, OpenAI)
- Refactor Assistant to use LLM-specific adapters
- Add base adapter class for common functionality


andreibondarev:
@khaile This is great, thank you. I renamed `open_ai.rb` to `openai.rb` because that's what the other file was named.

I'm not a fan of the `_base.rb` as the file name but I see why it's needed. I'd rather required the files explicitly then.

I'll merge for now. Thank you!

PR #811, opened 2024-10-04
Support AWS Bedrock for assistant

khaile:
Leverage the Anthropic adapter, since `AwsBedrock#chat` only supports Anthropic.

Docs: https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html

TODO: Implement Converse API for AWS Bedrock LLM

> We recommend that you use the Converse API for integrating tool use into your application. For more information, see [Use a tool to complete an Amazon Bedrock model response](https://docs.aws.amazon.com/bedrock/latest/userguide/tool-use.html).


lib/langchain/assistants/llm/adapters/aws_bedrock.rb
```diff
@@ -0,0 +1,10 @@
+module Langchain
+  class Assistant
+    module LLM
+      module Adapters
+        class AwsBedrock < Anthropic
```
andreibondarev, line 5: Is this adapter only going to work with the Anthropic models hosted on AWS Bedrock? Aren't there other chat models that are accessible on AWS Bedrock?

khaile: Yeah, there are numerous chat models available on AWS Bedrock. However, the `LLM::AwsBedrock#chat` currently only supports the Anthropic models. I attempted to use the `switch...case` approach to support models other than the Anthropic ones, but it isn't feasible ATM.

In the future, we may consider implementing the `LLM::AwsBedrock#converse` to accommodate all chat models for the assistant instead of the `LLM::AwsBedrock#chat` (invoke_model).

djpate:
Would need something like this for my use case too.

andreibondarev:
> Would need something like this for my use case too.

@djpate Which model are you planning on using via AWS Bedrock?

djpate:
> > Would need something like this for my use case too.
> 
> @djpate Which model are you planning on using via AWS Bedrock?

@andreibondarev I'm using `anthropic.claude-3-5-sonnet-20240620-v1:0'` in my current custom implementation. Now that we are trying to productionize it I'd rather use a library like this one but can't switch just yet.

andreibondarev:
> > > Would need something like this for my use case too.
> > 
> > 
> > @djpate Which model are you planning on using via AWS Bedrock?
> 
> @andreibondarev I'm using `anthropic.claude-3-5-sonnet-20240620-v1:0'` in my current custom implementation. Now that we are trying to productionize it I'd rather use a library like this one but can't switch just yet.

@djpate Mind taking [this](https://github.com/patterns-ai-core/langchainrb/pull/849) for a quick spin and see if it works as expected?

andreibondarev:
@khaile Please look over my PR: https://github.com/patterns-ai-core/langchainrb/pull/849. If you could test it out -- I'd be happy to drive it across the finish line.

khaile:
> @khaile Please look over my PR: #849. If you could test it out -- I'd be happy to drive it across the finish line.

@andreibondarev Thank you for this! I tested it, and it works well 👍🏽

PR #818, opened 2024-10-09
Update README.md with gemini chat()

palladius:
Documenting ruby invocation for `chat()`
as it took me a while to make it work (unlike embeddings which was a no brainer)


andreibondarev:
@palladius I went ahead and made a more subtle note in the README [here](https://github.com/patterns-ai-core/langchainrb/commit/1b8338a108bd8c125b86a86814b632d0dd868f1d). Thank you!

PR #836, opened 2024-10-14
Fix issue with clearing instructions

qarol:
Previously, when instructions were assigned, setting them to `nil` did not fully clear the value. This commit resolves that issue.

### DIFF
--- a/lib/langchain/assistant.rb
+++ b/lib/langchain/assistant.rb
@@ -196,7 +196,7 @@ module Langchain
 
       if @llm_adapter.support_system_message?
         # TODO: Should we still set a system message even if @instructions is "" or nil?
-        replace_system_message!(content: new_instructions) if @instructions
+        replace_system_message!(content: new_instructions)
       end
     end
 
@@ -217,6 +217,7 @@ module Langchain
     # @return [Array<Langchain::Message>] The messages
     def replace_system_message!(content:)
       messages.delete_if(&:system?)
+      return if content.nil?
 
       message = build_message(role: "system", content: content)
       messages.unshift(message)
--- a/spec/langchain/assistant/assistant_spec.rb
+++ b/spec/langchain/assistant/assistant_spec.rb
@@ -482,6 +482,12 @@ RSpec.describe Langchain::Assistant do
         expect(subject.messages.first.content).to eq("New instructions")
         expect(subject.instructions).to eq("New instructions")
       end
+
+      it "clears instructions when instructions are nil" do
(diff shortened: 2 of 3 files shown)


lib/langchain/assistant.rb
```diff
@@ -217,6 +217,7 @@ def tool_choice=(new_tool_choice)
     # @return [Array<Langchain::Message>] The messages
     def replace_system_message!(content:)
       messages.delete_if(&:system?)
+      return if content.nil?
```
andreibondarev, line 220: This makes sure we avoid setting a message with nil content, right?

lib/langchain/assistant.rb
```diff
@@ -196,7 +196,7 @@ def instructions=(new_instructions)
 
       if @llm_adapter.support_system_message?
         # TODO: Should we still set a system message even if @instructions is "" or nil?
```
andreibondarev, line 198: I think this comment can be deleted?

qarol: That comment is still valid, you can set "" and it will be used as system message. You can send empty string to OpenAI and the LLM will accept that. Hard to say if it brings any value to the conversation.
That PR solves one issue, when you had a non-empty system message but decides to clear it using `nil` the old instruction stays unchanged in messages list. Affected LLMs that supports system message only.

lib/langchain/assistant.rb
qarol: Exactly, there is no point to keep message with nil content.

spec/langchain/assistant/assistant_spec.rb
```diff
@@ -482,6 +482,12 @@
         expect(subject.messages.first.content).to eq("New instructions")
         expect(subject.instructions).to eq("New instructions")
       end
+
+      it "clears instructions when instructions are nil" do
+        subject.instructions = nil
+        expect(subject.messages).to be_empty
+        expect(subject.instructions).to be_nil
```
andreibondarev, line 489: Can we enhance this spec with:
```ruby
expect(subject.messages.find { |m| m.system? }).to be_present
subject.instructions = nil
expect(subject.messages).to be_empty
expect(subject.instructions).to be_nil
expect(subject.messages.find { |m| m.system? }).to_not be_present
```

To make sure that the system message gets dropped.

andreibondarev:
@qarol Great job, thank you!

qarol:
Thanks @andreibondarev for taking care of those specs, good thoughts.

Issue #826, opened 2024-10-12
Anthropic throws an error when Database tool is used


andreibondarev:
**Describe the bug**

**To Reproduce**
```ruby
llm = Langchain::LLM::Anthropic.new(api_key: ENV["ANTHROPIC_API_KEY"])
llm.chat messages:[{role:"user", content:"hello"}], tools: Langchain::Tool::Database.function_schemas.to_anthropic_format
```

**Expected behavior**
The method call is not supposed to throw an error.

**Desktop (please complete the following information):**
- Ruby version: 3.3.0
- Langchain.rb version: 0.17.1

andreibondarev:
Trace:
```
irb(main):006> llm.chat messages:[{role:"user", content:"hello"}], tools: Langchain::Tool::Database.new(connection_string: "postgres://localhost:5432/db_name")

I, [2024-10-11T20:51:26.344119 #62514]  INFO -- request: POST https://api.anthropic.com/v1/messages
I, [2024-10-11T20:51:26.344183 #62514]  INFO -- request: x-api-key: "[MASKED]"
anthropic-version: "2023-06-01"
Content-Type: "application/json"
I, [2024-10-11T20:51:26.344214 #62514]  INFO -- request: {"messages":[{"role":"user","content":"hello"}],"tools":"#<Langchain::Tool::Database:0x0000000123294310>","model":"claude-3-5-sonnet-20240620","max_tokens":256,"temperature":0.0}
I, [2024-10-11T20:51:26.674608 #62514]  INFO -- error: /Users/andrei/.asdf/installs/ruby/3.3.0/lib/ruby/gems/3.3.0/gems/faraday-2.12.0/lib/faraday/response/raise_error.rb:30:in `on_complete': the server responded with status 400 (Faraday::BadRequestError)
	from /Users/andrei/.asdf/installs/ruby/3.3.0/lib/ruby/gems/3.3.0/gems/faraday-2.12.0/lib/faraday/middleware.rb:57:in `block in call'
	from /Users/andrei/.asdf/installs/ruby/3.3.0/lib/ruby/gems/3.3.0/gems/faraday-2.12.0/lib/faraday/response.rb:42:in `on_complete'
	from /Users/andrei/.asdf/installs/ruby/3.3.0/lib/ruby/gems/3.3.0/gems/faraday-2.12.0/lib/faraday/middleware.rb:56:in `call'
[...]

PR #837, opened 2024-10-14
Anthropic Tools: added default `input_schema` if no params

dghirardo:
### Issue #826

This pull request addresses the issue where function schemas without parameters were causing errors when used with Anthropic. It ensures that a default empty `input_schema` is added when no parameters are present, aligning with [Anthropic's requirements](https://docs.anthropic.com/en/docs/build-with-claude/tool-use).


andreibondarev:
@dghirardo Amazing work! 👏🏻

PR #840, opened 2024-10-15
Shorter model name params

### LINKED ISSUE #820: Rename the model name related parameter names  to shorter versions
When instantiating the LLM-related constructors it is quite long to type "default_options: { chat_completion_model_name: }". We should rename it to "chat_model:", short and sweet.

bborn:
Fixes #820:
- renames `chat_completion_model_name` to `chat_model`
- renames `completion_model_name` to `completion_model`
- renames `embeddings_model_name` to `embedding_model`

**This is a breaking change**, and I updated the changelog and bumped the version to reflect that.


qarol:
What do you think to release new versions following Semantic Versioning https://semver.org? If the release introduces breaking change, then bumping major version should help to catch potential problems earlier.

bborn:
@qarol  you're probably right - a breaking change should be a MAJOR version change according to server. @andreibondarev  I'll leave that decision to you though...

Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    langchainrb (0.18.0)
+    langchainrb (0.19.0)
```
andreibondarev, line 4: Mind reverting this? I'll bump this up when I release a new version.

lib/langchain/llm/aws_bedrock.rb
```diff
@@ -11,8 +11,8 @@ module Langchain::LLM
   #
   class AwsBedrock < Base
     DEFAULTS = {
-      chat_completion_model_name: "anthropic.claude-v2",
-      completion_model_name: "anthropic.claude-v2",
+      chat_model: "anthropic.claude-v2",
+      completion_model: "anthropic.claude-v2",
       embeddings_model_name: "amazon.titan-embed-text-v1",
```
andreibondarev, line 16: I'm thinking maybe we rename this as well, to `embedding_model`?

andreibondarev:
@bborn Quick thought -- maybe since the 3 methods are called:
```ruby
def embed
def complete
def chat
```
we should call the defaults:
```
:embed_model
:complete_model
:chat_model
```
What do you think?

CHANGELOG.md
```diff
@@ -25,56 +33,67 @@
...
 - [BREAKING] Removing Langchain::LLM::GooglePalm
 - Assistant can now process image_urls in the messages (currently only for OpenAI and Mistral AI)
 - Vectorsearch providers utilize the global Langchain.logger
 - Update required milvus, qdrant and weaviate versions
 
 ## [0.16.1] - 2024-09-30
+
 - Deprecate Langchain::LLM::GooglePalm
-- Allow setting response_object: {} parameter when initializing supported Langchain::LLM::* classes
+- Allow setting response_object: {} parameter when initializing supported Langchain::LLM::\* classes
```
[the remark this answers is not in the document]
bborn: Weird - I'm not sure why that's there. I think Prettier added it or something. Removing ...

bborn:
@andreibondarev I think that makes sense. Updated. 

My only concern is that `complete_model` just sounds kind of ... weird to me. But it's consistent with the method names so probably ok.

andreibondarev:
> What do you think to release new versions following Semantic Versioning https://semver.org? If the release introduces breaking change, then bumping major version should help to catch potential problems earlier.

I love that! Given that we're pre-1.0.0, do you suggest that we keep incrementing minor versions, 0.18.0, 0.19.0, 0.20.0, etc.?

andreibondarev:
@bborn Great PR! Thank you!

qarol:
> > What do you think to release new versions following Semantic Versioning https://semver.org? If the release introduces breaking change, then bumping major version should help to catch potential problems earlier.
> 
> @qarol I love that! Given that we're pre-1.0.0, do you suggest that we keep incrementing minor versions, 0.18.0, 0.19.0, 0.20.0, etc.?

@andreibondarev question is, what things blocks from releasing 1.0.0 ;) For pre-1.0.0 it's acceptable to bump just minor version no matter if it's a breaking change or new feature. For bugfixes increasing patch number should be enough IMO

PR #850, opened 2024-10-20
Fix MistralAIMessage to handle "Tool" Output 

dkaplan88:
Solves: https://github.com/patterns-ai-core/langchainrb/issues/839

Breaks up hash by specific role following API Conventions

https://docs.mistral.ai/api/#tag/chat/operation/chat_completion_v1_chat_completions_post
Assistant
image

System
image

Tool
image

User
image

### DIFF
--- a/lib/langchain/assistant/messages/mistral_ai_message.rb
+++ b/lib/langchain/assistant/messages/mistral_ai_message.rb
@@ -45,11 +45,14 @@ module Langchain
         #
         # @return [Hash] The message as an MistralAI API-compatible hash
         def to_hash
-          {}.tap do |h|
-            to_system_hash(h)    if system?
-            to_assistant_hash(h) if assistant?
-            to_user_hash(h)      if user?
-            to_tool_hash(h)      if tool?
+          if assistant?
+            assistant_hash
+          elsif system?
+            system_hash
+          elsif tool?
+            tool_hash
+          elsif user?
+            user_hash
           end
         end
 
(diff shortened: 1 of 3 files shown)


lib/langchain/assistant/messages/mistral_ai_message.rb
```diff
@@ -46,29 +46,10 @@ def llm?
         # @return [Hash] The message as an MistralAI API-compatible hash
         def to_hash
```
andreibondarev, line 47: What're your thoughts on doing something like this:
```ruby
def to_hash
  if assistant?
    to_assistant_hash
  elsif tool?
    to_tool_hash
  ...
end

def to_assistant_hash
  h = {
    role: role,
    content: to_content
  }
  h[:tool_calls] = tool_calls if tool_calls.any?
end
```

andreibondarev: It's pretty similar to what you've got just slightly different stylistically.

dkaplan88: Hmm, I am very on the fence about it as one hash or hash by role. 

I think declaring the separation by role makes sense since it feels like Mistral has declared separate API Contracts by role (as we see with `content`) 😄 

I updated it to do it by role and refactored the content hash so it was just building the array value.

andreibondarev:
@dkaplan88 Thank you! Great work!

PR #851, opened 2024-10-21
Improvements to Evals

bborn:
Just exploring some ideas:

- Should be able to run evals against a full dataset (instead of one item at a time)
- Should be able to run multiple evaluators at once
- Add a Regex evaluator
- Add cosine similarity evaluator
- Add a LLM evaluator

### DIFF
--- a/lib/langchain/evals/base.rb
+++ b/lib/langchain/evals/base.rb
@@ -21,7 +21,7 @@ module Langchain
           dataset_item_scores = {}
 
           evaluators.each do |evaluator|
-            dataset_item_scores[evaluator.class.name.split("::").last] = evaluator.score(question: data[:question], answer: data[:answer], context: data[:context])
+            dataset_item_scores[evaluator.class.name.split("::").last] = evaluator.score(**data)
           end
 
           scored_dataset << {
--- a/lib/langchain/evals/base.rb
+++ b/lib/langchain/evals/base.rb
@@ -1,0 +1,36 @@
+module Langchain
+  module Evals
+    class Base
+      # Evaluates a dataset with multiple evaluators
+      # {
+      #   question: "question",
+      #   answer: "answer",
+      #   context: "Context"
+      #   "AnswerRelevance" => Float,
+      #   "ContextRelevance" => Float,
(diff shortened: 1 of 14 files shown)


andreibondarev:
@bborn How useful is the regex evaluator in your opinion? 

Do you have any thoughts evals that calculate vector distance?

bborn:
@andreibondarev I think regex or other string comparison is pretty important. You might want to ensure your model/agent is returning a number, a URL, an email, etc. Or you might want to ensure that the answer contains the expected output string (this doesn't exist in Lanchain.rb yet), something like:

```
regex_evaluator.score(answer: "The answer to 2 + 2 is 4", expected_answer: "4")
```

Vector (or levenshtein, etc.) distance seems useful too. Not so much as an absolute score but as something you could look at over time (if our agent was getting a vector score of .75 for the last three months, and then we changed the prompt and now it's getting .45, we'd be concerned). 


I think the evaluators kind of break down into LLM Graded, LLM Labeled, and Code Graded):

CleanShot 2024-10-21 at 08 21 04@2x

LLM Graded: ask another LLM to score the dataset item based on some criteria
LLM Labled: ask an LLM to label the dataset item
Code Graded: run the dataset item through some grading algorithm (regex, JSON, or other)

bborn:
Another thought: maybe you should be able to add an Eval to your Agent or llm call like this:

```
dataset = "/path/to/dataset.jsonl"

evaluators = [
    Langchain::Evals::LLM::LLM.new(llm: llm),
    Langchain::Evals::LLM::CosineSimilarity.new(llm: llm)
]

response = llm.complete(prompt: "Once upon a time", evaluate_with: {evaluators: evaluators, dataset: dataset} )
```

sergiobayona:
Thanks for the exploratory work on evaluators!
We agree evals belong in the ecosystem, but we need a broader discussion about whether they should live inside LangChain-RB or in a companion gem.
Because this PR is still marked Draft and has unresolved review comments, I’m closing it for now.
When we’ve decided on the best home for evals—or if you’d like to split this into smaller, ready-for-review PRs—feel free to reopen or submit a new one. 🙏

PR #872, opened 2024-11-12
Replace auto_tool_execution with execute_tools

mattlindsey:
WIP

Note: Also changes current default behavior of `assistant.run()` of not automatically running tools.


andreibondarev:
@mattlindsey Do you think you could separate renaming the parameter to a separate smaller PR? Thanks!

PR #873, opened 2024-11-12
Replace auto_tool_execution with execute_tools (just step 1)

mattlindsey:
WIP

@andreibondarev Can you give this a try and see if it's the behavior you want.  Then I can fix the tests.

### DIFF
--- a/lib/langchain/assistant.rb
+++ b/lib/langchain/assistant.rb
@@ -138,8 +138,12 @@ module Langchain
         return
       end
 
-      @state = :in_progress
-      @state = handle_state until run_finished?(execute_tools)
+      if !execute_tools
+        @state = :completed
+      else
+        @state = :in_progress
+        @state = handle_state until run_finished?(execute_tools)
+      end
 
       messages
     end
--- a/lib/langchain/assistant.rb
+++ b/lib/langchain/assistant.rb
@@ -129,9 +129,9 @@ module Langchain
 
     # Run the assistant
     #
-    # @param auto_tool_execution [Boolean] Whether or not to automatically run tools
+    # @param execute_tools [Boolean] Whether or not to automatically run tools
     # @return [Array<Langchain::Message>] The messages
-    def run(auto_tool_execution: false)
+    def run(execute_tools: false)
       if messages.empty?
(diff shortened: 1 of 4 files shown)


lib/langchain/assistant.rb
```diff
@@ -129,17 +129,21 @@ def add_messages(messages:)
...
       if messages.empty?
         Langchain.logger.warn("#{self.class} - No messages to process")
         @state = :completed
         return
       end
 
-      @state = :in_progress
-      @state = handle_state until run_finished?(auto_tool_execution)
+      if !execute_tools
+        @state = :completed
```
mattlindsey, line 142: This is the part that confuses me.  Why are we even calling `run` if we don't want to do anything?
@andreibondarev If you want to handle this issue yourself, that's totally fine.  I don't want to clutter up the code.

andreibondarev: Why do we need to change the logic if we're merely renaming things?

mattlindsey: There's already a method called `execute_tools` so it's not that simple.

mattlindsey: For some reason, doing that I get the error below, then I get lost in the details.  I'm probably missing something simple.

OpenAI HTTP Error (spotted in ruby-openai 7.1.0): {"error"=>{"message"=>"An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages: call_ZM6cJs8AMnU0JcEdFErekxlj", "type"=>"invalid_request_error", "param"=>"messages.[3].role", "code"=>nil}}

mattlindsey: Oh wait.  You get that error also with the current code if you do this sequence, so it's not an issue with the rename:

a.add_message_and_run(content: "What's the weather in Boston?", auto_tool_execution: false)
a.add_message_and_run(content: "What's the weather in Chicago?", auto_tool_execution: true)

That is confusing.  Wasn't there something in the README before about running tools manually?  Should that sequence really cause an error?

I'll close these PRs and can do the simple rename later today or tomorrow when hopefully it will make more sense to me. LOL.

Issue #829, opened 2024-10-12
Introduce `assistant.execute_tools = true/false` option to replace the `auto_tool_execution` parameter.


andreibondarev:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #874, opened 2024-11-12
Replace auto_tool_execution with execute_tools (just first step) 

mattlindsey:
First step of issue #829


lib/langchain/assistant.rb
```diff
@@ -129,17 +129,17 @@ def add_messages(messages:)
 
     # Run the assistant
     #
-    # @param auto_tool_execution [Boolean] Whether or not to automatically run tools
+    # @param execute_tools [Boolean] Whether or not to automatically run tools
     # @return [Array<Langchain::Message>] The messages
-    def run(auto_tool_execution: false)
+    def run(execute_tools: false)
```
koic, line 134: A simple renaming would introduce a breaking change. A more user-friendly migration would likely be to issue a deprecation warning when the old option name (`:auto_tool_execution`) is used. Using the new option name  (`:execute_tools`) does not produce any warnings.

According to #829, it also indicates that the default value of these options will change from `false` to `true`. It would be best for #829 to deliver this as a single change to users, rather than breaking it into separate steps.

PR #889, opened 2024-11-30
Default Ollama to use llama3.2

WToa:
The quickstart instructions in the [ollama repo](https://github.com/ollama/ollama?tab=readme-ov-file#quickstart) now point to running:

```ollama run llama3.2```

Updating our ollama class to also use llama3.2 as the default.


WToa:
Are there any tests that require updating @andreibondarev ? I can see that we have a couple of tests that use llama3.1 but not sure if it's required to update those since they default to using the 3.1 model as part of VCR

PR #891, opened 2024-12-01
Langchain::Tool::Database#describe_table with COMMENT

Bahanix:
Allows LLM to make use of columns' comments.
Possible since Sequel 5.87.0 ([PR](https://github.com/jeremyevans/sequel/pull/2249)).

## Example of usage:

Given this structure:

```sql
CREATE TABLE users(
  id integer PRIMARY KEY,
  name string COMMENT 'Internally known as 'foobar''
);
```

I can know ask things like `Give me the last registered user's foobar` to the LLM.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -382,6 +382,7 @@ GEM
     safe_ruby (1.0.4)
       childprocess (>= 0.3.9)
     sequel (5.87.0)
+      bigdecimal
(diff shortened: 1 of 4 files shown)


andreibondarev:
@Bahanix Thank you for the PR. Pretty sure you need to update the version in the gemspec file as well.

PR #900, opened 2024-12-16
Empty text content should not be set when content is nil when using AnthropicMessage

andreibondarev:
…nthropicMessage


ms-ati:
Wow! Thank you so much @andreibondarev. I will run a few more tests, but on first run this did appear to fix the issue!

ms-ati:
Confirmed @andreibondarev this appears to have fixed it! TYSM

sins921 (approved):

PR #906, opened 2025-01-15
Add sqlite-vec support

CarlQLange:
This PR adds some support for [sqlite-vec](https://github.com/asg017/sqlite-vec).

Unfortunately I am brand new to Ruby so there was plenty of help from professor ChatGPT, so I would strongly recommend a close inspection to make sure I got things right. Having said that, the example script appears to work on my machine. I hope if nothing else that this may be a useful starting point to add SQLite support to this library.

### DIFF
--- a/lib/langchain/vectorsearch/sqlite_vec.rb
+++ b/lib/langchain/vectorsearch/sqlite_vec.rb
@@ -1,0 +1,154 @@
+# frozen_string_literal: true
+
+require "sqlite_vec"
+module Langchain::Vectorsearch
+  class SqliteVec < Base
+    #
+    # The SQLite vector search adapter using sqlite-vec
+    #
+    # Gem requirements:
+    #     gem "sqlite3", "~> 2.5"
+    #     gem "sqlite_vec", "~> 0.16.0"
+    #
+    # Usage:
+    #     sqlite_vec = Langchain::Vectorsearch::SqliteVec.new(url:, index_name:, llm:, namespace: nil)
+    #
+
+    attr_reader :db, :table_name, :namespace_column, :namespace
+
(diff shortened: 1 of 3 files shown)


lib/langchain/dependency_helper.rb
```diff
@@ -15,7 +15,12 @@ class VersionError < ScriptError; end
     # @raise [VersionError] If the gem is installed, but the version does not meet the requirements
     #
     def depends_on(gem_name, req: true)
-      gem(gem_name) # require the gem
+      if gem_name == "sqlite_vec"
```
andreibondarev, line 18: @CarlQLange Why is this if/else statement needed? Isn't the gem version what gets required?

CarlQLange: Woof, I have no memory of this. I think it was something to do with using the gem locally from something else. Sorry, I pretty much have no idea what I was doing back then!

lib/langchain/vectorsearch/sqlite_vec.rb
```diff
@@ -0,0 +1,154 @@
...
+      @db.execute("DROP TABLE IF EXISTS #{table_name}")
+    end
+
+    # Add a list of texts to the index
+    # @param texts [Array<String>] The texts to add to the index
+    # @param ids [Array<String>] The ids to add to the index, in the same order as the texts
+    # @return [Array<Integer>] The ids of the added texts
+    def add_texts(texts:, ids: nil)
+      if ids.nil? || ids.empty?
+        max_rowid = @db.execute("SELECT MAX(rowid) FROM #{table_name}").first.first || 0
```
andreibondarev, line 59: Doesn't the `rowid` auto increment?

CarlQLange:
Closing this PR - someone with more ability and experience could probably do a better job.

PR #917, opened 2025-02-08
GigaChat LLM support

skk2010:
Implemented GigaChat LLM support. Please take a look.


skk2010:
Not needed. Use https://github.com/ai-forever/gpt2giga

Issue #909, opened 2025-01-23
Cannot exclude temperature param (it is unsupported by OpenAI o1)


sansari:
**Describe the bug**
When using OpenAI's o1 model (e.g. `o1-2024-12-17`), the `temperature` param is not supported. However, if you want to use the `chat` method and pass no temperature, it's not currently possible, due to the fallback to `DEFAULTS`:

![Image](https://github.com/user-attachments/assets/24f71fcd-4e23-4cc6-9033-72181a845e72)

**To Reproduce**
```
chat(messages: [...], model: "o1-2024-12-17", temperature: nil)
chat(messages: [...], model: "o1-2024-12-17")
```

Both of the above will end up using the `DEFAULTS` temperature value which is setup inside the internals of the library and cannot be easily overridden.

**Expected behavior**
We probably want something similar to this in how it removes the dimension parameter for a particular model: https://github.com/patterns-ai-core/langchainrb/blob/main/lib/langchain/llm/openai.rb#L86

**Screenshots**

![Image](https://github.com/user-attachments/assets/f689afa2-9ddf-4d14-aa5d-82e6772be7fb)

**Desktop (please complete the following information):**

- OS: Linux
- Ruby version 3.3.5
- Langchain.rb version 0.19.2

manoelneto:
Insanely how no one touched this yet =x I'm still waiting for this.

cmer:
Have you tried: `new(api_key: "abc", llm_options: { model: "o1-2024-12-17" }, default_options: { temperature: nil })`?

manoelneto:
> Have you tried the following?
> 
> new(api_key: "abc", llm_options: { model: "o1-2024-12-17" }, default_options: { temperature: nil })

You're a genius, that worked to me

```
client = Langchain::LLM::OpenAI.new(
  api_key: OpenAiClient::API_KEY,
  default_options: {
    temperature: nil,
    chat_model: 'o3-mini'
  }
)


client.chat(messages: [{ role: :user, content: 'Hello, how are you?' }])
```

ajsharp:
Agree here -- the reasoning models don't support the `temperature` param, so this library should not add it by default.

PR #921, opened 2025-02-14
Allow o3-mini usage with tools

almathie:
OpenAI o3-mini model is fully compatible with tool calls, but selecting `chat_model: 'o3-mini'` and providing tools in a langchainrb assistant fails.

The error returned by OpenAI specifies that the `parallel_tool_calls` parameter is unsupported by the `o3` family models.
The error is a bit similar to: https://github.com/patterns-ai-core/langchainrb/issues/909 but instead of being about the `temperature` parameter, this one is about the `parallel_tool_calls`

On issue #909 users are able to circumvent the error by explicitly setting the `temperature` parameter to `nil`.
This workaround does not work for the `parallel_tool_calls` parameter: if a users sets it to `nil`, the parameter is still sent to OpenAI (with a `null` value) and OpenAI complains that `null` is not an allowed value for a boolean.

This PR changes the behavior so that if `parallel_tool_calls` parameter is set to `nil` no value is sent for it to OpenAI, mimicking the behavior of the `temperature` parameter.

This allows users to use the `o3` family model with tools successfully; as long as they also set the `parallel_tool_calls` parameter to `nil`:

```Ruby
openai = Langchain::LLM::OpenAI.new(
  api_key: ENV["OPENAI_API_KEY"],
  default_options: {
    temperature: nil,
    chat_model: "o3-mini",
  }
)
assistant = Langchain::Assistant.new(
  llm: openai,
  instructions: "You are a simple assistant that answers user questions",
  parallel_tool_calls: nil, # <------ Required to be nil for o3 family models
  tools: [
    Langchain::Tool::Wikipedia.new,
  ]
)
```

### DIFF
--- a/lib/langchain/assistant/llm/adapters/openai.rb
+++ b/lib/langchain/assistant/llm/adapters/openai.rb
@@ -25,6 +25,7 @@ module Langchain
               params[:tools] = build_tools(tools)
               params[:tool_choice] = build_tool_choice(tool_choice)
               # Temporary fix because OpenAI o1/o3/reasoning models don't support `parallel_tool_calls` parameter.
+              # Set `Assistant.new(parallel_tool_calls: nil, ...)` to avoid the error.
               params[:parallel_tool_calls] = parallel_tool_calls unless parallel_tool_calls.nil?
             end
             params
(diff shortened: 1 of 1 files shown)


andreibondarev:
@almathie Thank you for the fix!

Issue #928, opened 2025-02-27
Support for Amazon Nova models


tpaulshippy:
**Is your feature request related to a problem? Please describe.**
Amazon Nova models were recently released and we would like to be able to use them.
https://docs.aws.amazon.com/nova/latest/userguide/what-is-nova.html

**Describe the solution you'd like**
Support for Nova inside the Bedrock client.

**Describe alternatives you've considered**
Using other models.

**Additional context**
These are supported in the python version of langchain. See https://github.com/langchain-ai/langchain-aws/blob/main/samples/models/getting_started_with_nova.ipynb

PR #929, opened 2025-02-27
Add support for Amazon Nova models

tpaulshippy:
## Purpose
Addresses #928 

## Approach
Add support for Amazon models in AwsBedrock and introduce AwsBedrockAmazonResponse.

## Testing
Ran against Nova Micro.
RSpec


mtrefilek:
@andreibondarev this is looking good to me. Are we able to get test runs going?

bak-luka:
Any progress on this? We have some usecases for Nova models and would like to use them with langchain.

PR #950, opened 2025-04-10
add support for distance filtering in vector search for pinecone

sergiobayona:


sergiobayona:
closing while we figure out a unified distance filtering api across providers.

PR #951, opened 2025-04-12
Support ruby-anthropic gem

koic:
This PR adds support for ruby-anthropic gem, which was renamed from anthropic gem.

> This gem has been renamed from `anthropic` to `ruby-anthropic`, to make way for the new [official Anthropic Ruby SDK](https://github.com/anthropics/anthropic-sdk-ruby) :tada: If you wish to keep using this gem, you just need to update your Gemfile from `anthropic` to `ruby-anthropic` and version `0.4.2` or greater. No other changes are needed and it will continue to work as normal.

https://github.com/alexrudall/ruby-anthropic/blob/v0.4.2/README.md

So, the anthropic gem was renamed to ruby-anthropic starting from version 0.4.2. If user is still using version 0.4.1 or earlier of the anthropic gem, the implementation falls back to anthropic instead of treating it as a load error.

The development dependency has been updated to use the new ruby-anthropic gem instead of the old anthropic gem.


sergiobayona:
👍🏼

sergiobayona (approved):

PR #953, opened 2025-04-14
fix: Handle nil response in OpenAI LLM streaming

sergiobayona:
When using the `chat` method of `Langchain::LLM::OpenAI` with a block for streaming responses, the underlying `ruby-openai` client returns `nil` after the stream completes.

The `with_api_error_handling` method previously attempted to call `.empty?` on this `nil` response, resulting in a `NoMethodError`.

This commit fixes the issue by adding a `response.nil? ||` check before calling `.empty?` in `with_api_error_handling`. This prevents the error when the client yields control back after finishing the stream.

Adds a new RSpec test case specifically for the streaming scenario. The test mocks the client to return `nil` after streaming and verifies that no error is raised and the final response object is correctly assembled from the collected chunks.

### DIFF
--- a/lib/langchain/llm/openai.rb
+++ b/lib/langchain/llm/openai.rb
@@ -173,7 +173,7 @@ module Langchain::LLM
 
     def with_api_error_handling
       response = yield
-      return if response.empty?
+      return if response.nil? || response.empty?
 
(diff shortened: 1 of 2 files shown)


andreibondarev:
@sergiobayona Very nice!

PR #955, opened 2025-04-14
feat: Add support for Anthropic 'thinking' parameter

sergiobayona:
This commit introduces support for the `thinking` parameter in the Anthropic LLM integration. Users can now enable Claude 3.7 Sonnet's extended thinking mode by passing the `thinking` parameter either via `default_options` during initialization or directly to the `chat` method.

Changes include:
- Added `thinking` to the `chat_parameters` schema in `Langchain::LLM::Anthropic`.
- Included tests for the `thinking` parameter in `spec/langchain/llm/anthropic_spec.rb`.
- Updated inline documentation for the `initialize` and `chat` methods to reflect the new parameter.

### DIFF
--- a/lib/langchain/llm/anthropic.rb
+++ b/lib/langchain/llm/anthropic.rb
@@ -22,7 +22,7 @@ module Langchain::LLM
     #
     # @param api_key [String] The API key to use
     # @param llm_options [Hash] Options to pass to the Anthropic client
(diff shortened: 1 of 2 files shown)


sergiobayona:
@andreibondarev noticing there is not a good place to document this feature at the moment. The readme does not itemize the models supported and their features.

sergiobayona:
There is an interaction between `max_tokens` parameter and the thinking's `budget_tokens` parameter where the max_tokens must be greater than the thinking's budget_tokens. This is not enforced in the code at the moment. Currently the Anthropic API will return an error unless this is the case.

PR #960, opened 2025-04-20
fix(Anthropic): Handle empty tool input

TheBerg:
When streaming a response, if a tool parameter is defined then JSON.parse breaks with an error because it is an empty string.


sergiobayona:
👍🏼

PR #976, opened 2025-04-30
Configure Linter for minimum Ruby version compatibility

koic:
Standard Ruby runs the linter using the runtime Ruby version by default, but developers may use any Ruby version supported by Langchain.rb. This means the linter should work correctly across all supported Ruby versions (3.1+), so it's best to specify the minimum supported version via the `ruby_version: 3.1` setting: https://github.com/standardrb/standard?tab=readme-ov-file#configuring-ruby_version

For reference, the Ruby version used in GitHub Actions for linting is currently set to 3.2, so the linter is effectively running under Ruby 3.2 at the moment: https://github.com/patterns-ai-core/langchainrb/blob/0.19.4/.github/workflows/ci.yml#L56

Since the `ruby_version: 3.1` setting takes precedence over the runtime version, this patch would also address that concern.


andreibondarev:
@koic Thank you! Tons of really great changes!

PR #980, opened 2025-05-02
Suppress a Ruby 3.4 warning for URI parser

koic:
This PR suppresses the following Ruby 3.4's warning. e.g.;

```console
$ ruby -v
ruby 3.4.3 (2025-04-14 revision d0b7e5b6a0) +PRISM [x86_64-darwin24]
$ RUBYOPT=-w bundle exec rspec spec/langchain/loader_spec.rb
/Users/koic/src/github.com/patterns-ai-core/langchainrb/lib/langchain/loader.rb:94:
warning: URI::RFC3986_PARSER.escape is obsolete. Use URI::RFC2396_PARSER.escape explicitly.
```

Compatibility with Ruby 3.1 supported by Langchain.rb is preserved.


lib/langchain/loader.rb
```diff
@@ -91,7 +91,11 @@ def load(&block)
 
     def load_from_url
       unescaped_url = URI.decode_www_form_component(@path)
-      escaped_url = URI::DEFAULT_PARSER.escape(unescaped_url)
+
+      # Ensure compatibility with `URI::RFC2396_PARSER` (Ruby 3.4+) and `URI::DEFAULT_PARSER` (<= 3.3).
+      uri_parser = defined?(URI::RFC2396_PARSER) ? URI::RFC2396_PARSER : URI::DEFAULT_PARSER
```
andreibondarev, line 96: Instead of doing this in a few different places -- it would probably make sense to define it as a Util and obfuscate the actual class used: `URI::RFC2396_PARSER` or `URI::DEFAULT_PARSER`?

koic: That makes sense! I opened PR #994.

PR #991, opened 2025-05-14
[Doc] Document deprecated APIs using `@deprecated` tag

koic:
This change adds the `@deprecated` YARD tag to several classes and methods.

This makes it possible to explicitly mark the API as deprecated in the YARD documentation. It also updates inline references from backtick-style (`name`) to the link-style (`{name}`) for improved documentation rendering.

langchain-llm-llamacpp


andreibondarev:
@koic Wow, I never knew about this!

PR #996, opened 2025-05-18
Use `public_send` instead of `send`

koic:
Since tool invocation from assistant is expected to be limited to public methods, `public_send` is used instead of `send` to avoid unintended invocation of private method.


sergiobayona:
👍🏼

PR #1003, opened 2025-06-06
 Convert LangchainRB to a Rails Engine

lemingos:
# Convert LangchainRB to a Rails Engine

## Description
This PR transforms LangchainRB from a plain Ruby gem into a full-fledged Rails::Engine, enabling tighter Rails integration and better support for Rails applications.

### 1. Added Rails Engine Structure
- Created `Langchain::Engine` class inheriting from `Rails::Engine`
- Set up proper autoloading and eager loading paths
- Isolated the engine namespace to prevent conflicts
- add Response namespace to classes located in `lib/langchain/llm/response` directory

### 2. Rails Integration
- Added `rails` as a runtime dependency in gemspec
- Set up test/dummy Rails application for integration testing

### 3. Testing Infrastructure
- Added `rspec-rails` for Rails-specific testing
- Added Rails-specific test helpers and configurations

## Benefits
- **Namespace isolation** all classas are now scoped to Langchain
- **Better Rails Integration**: Seamless integration with Rails applications
- **Easier Configuration**: Standard Rails initializers and configurations
- **Improved Testing**: Better test coverage with Rails integration tests
- **Future Extensibility**: Foundation for adding persistance layer

## Breaking Changes
- This change makes Rails a required dependency

## Testing
- [x] All existing tests pass


mvz:
This makes langchainrb depend on Rails. Is that intentional and what will the relation be with langchainrb_rails? I'm currently using langchainrb in contexts that do not involve Rails at all.

ebababi:
For what it's worth, we were also using `langchainrb` in contexts that don't involve Rails at all. As already mentioned, shouldn't this tighter Rails integration be part of `langchainrb_rails`?

Issue #1015, opened 2025-07-16
The output fixing parser isn't working with Gemini 2.5 flash.


carlows:
**Describe the bug**
The output fixing parser isn't working with Gemini 2.5 flash.

As you described in the readme, google models require a different format:

```
messages = [
  { role: "system", content: "You are a helpful assistant." },
  { role: "user", content: "What's the weather like today?" }
  # Google Gemini and Google VertexAI expect messages in a different format:
  # { role: "user", parts: [{ text: "why is the sky blue?" }]}
]
```

However, the OutputFixingParser does not take that into account: https://github.com/patterns-ai-core/langchainrb/blob/main/lib/langchain/output_parsers/output_fixing_parser.rb#L50-L55

**To Reproduce**
Using the output fixing parser:

```
parser = Langchain::OutputParsers::StructuredOutputParser.from_json_schema('some_schema.json'))
Langchain::OutputParsers::OutputFixingParser.from_llm(llm:, parser:)
```

Raises an error when using the model gemini 2.5-flash.

**Expected behavior**
The output fixing parser should work with google/gemini models

**Terminal commands & output**
Commands you used and the terminal output.

**Screenshots**

**Desktop (please complete the following information):**

- OS: [e.g. OS X, Linux, Ubuntu, Windows]
- Ruby version [3.3]
- Langchain.rb version [0.19.5]

**Additional context**

carlows:
The error from google:

> {"error"=>{"code"=>400, "message"=>"Invalid JSON payload received. Unknown name \"content\" at 'contents[0]': Cannot find field.", "status"=>"INVALID_ARGUMENT", "details"=>[{"@type"=>"type.googleapis.com/google.rpc.BadRequest", "fieldViolations"=>[{"field"=>"contents[0]", "description"=>"Invalid JSON payload received. Unknown name \"content\" at 'contents[0]': Cannot find field."}]}]}}

carlows:
Opened a PR with a suggestion, but maybe there's a way to do it with the UnifiedParameters? Or I'm missing something obvious...

PR #1016, opened 2025-07-16
Add an exception for GoogleGemini and GoogleVertexAI when using the OutputFixingParser

carlows:
Connected to #1015 

As described in the README, the google APIs have a different format for their LLMs:

```
messages = [
  { role: "system", content: "You are a helpful assistant." },
  { role: "user", content: "What's the weather like today?" }
  # Google Gemini and Google VertexAI expect messages in a different format:
  # { role: "user", parts: [{ text: "why is the sky blue?" }]}
]
response = llm.chat(messages: messages)
chat_completion = response.chat_completion
```

And the OutputFixingParser class doesn't work with Gemini 2.5-flash, as it throws the following error:

> {"error"=>{"code"=>400, "message"=>"Invalid JSON payload received. Unknown name "content" at 'contents[0]': Cannot find field.", "status"=>"INVALID_ARGUMENT", "details"=>[{"@type"=>"type.googleapis.com/google.rpc.BadRequest", "fieldViolations"=>[{"field"=>"contents[0]", "description"=>"Invalid JSON payload received. Unknown name "content" at 'contents[0]': Cannot find field."}]}]}}

The goal of this PR is to use the correct format instead.

### DIFF
--- a/lib/langchain/output_parsers/output_fixing_parser.rb
+++ b/lib/langchain/output_parsers/output_fixing_parser.rb
@@ -47,12 +47,7 @@ module Langchain::OutputParsers
       parser.parse(completion)
     rescue OutputParserException => e
       new_completion = llm.chat(
-        messages: [{role: "user",
-                    content: prompt.format(
-                      instructions: parser.get_format_instructions,
-                      completion: completion,
-                      error: e
-                    )}]
+        messages: chat_messages(completion, e)
       ).completion
       parser.parse(new_completion)
     end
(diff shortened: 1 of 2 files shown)


lib/langchain/output_parsers/output_fixing_parser.rb
```diff
@@ -47,12 +47,7 @@ def parse(completion)
...
     rescue OutputParserException => e
       new_completion = llm.chat(
-        messages: [{role: "user",
-                    content: prompt.format(
-                      instructions: parser.get_format_instructions,
-                      completion: completion,
-                      error: e
-                    )}]
+        messages: chat_messages(completion, e)
       ).completion
```
konzinov, line 51: the call to `.completion` here should be `.chat_completion` for Vertex and Gemini LLM clients.
Otherwise it raises `NotImplementedError: NotImplementedError`

Issue #1011, opened 2025-06-30
Assistant raise exceptions when OpenAI tool_calls is nil


littlechu:
**Describe the bug**
Seem like in OpenAIMessage can pass in tool_calls = nil. which causes exceptions being throw

**To Reproduce**
Create OpenAI assistant agent with something simple, like the following
   assistant = Langchain::Assistant.new(
      llm: openai,
      instructions: "You're a helpful AI assistant",
      tools: [Langchain::Tool::FileSystem.new, Langchain::Tool::Calculator.new]
    )
    assistant.add_message_and_run!(content: 'how many directory are there under /Home')
    this will result in something like this
             raise ArgumentError, "Tool calls must be an array of hashes" unless tool_calls.is_a?(Array) && tool_calls.all? { |tool_call| tool_call.is_a?(Hash) }

**Expected behavior**
I expect no exception is being raise.

- OS: [e.g. OS X, Linux, Ubuntu, Windows]
- Ruby version 3.4.2
- Langchain.rb version 0.19.5

**Additional context**
By patching the following two places, the code seem to be working
lib/langchain/assistant/messages/openai_message.rb:31 add tool_calls ||=[]
lib/langchain/assistant.rb:312 change response.tool_calls.any? to response.tool_calls&.any?

Seem like it's possible for the repose body to contain "tool_calls":null from openai gems

I, [2025-06-30T14:39:13.533775 #94841]  INFO -- [Langchain.rb]: response: {"id":"chatcmpl-BoGVVtq92F0jUwOIs6CU0qs65xbql","created":1751319553,"model":"gpt-4o-mini-2024-07-18","object":"chat.completion","system_fingerprint":"fp_efad92c60b","choices":[{"finish_reason":"stop","index":0,"message":{"content":"There are 2 entries in the `/Home` directory: `.` (current directory) and `..` (parent directory). It appears that there are no additional directories under `/Home`.","role":"assistant","tool_calls":null,"function_call":null,"annotations":[]},

PR #1017, opened 2025-07-17
Fix Assistant exception with OpenAI

littlechu:
OpenAI can report tool_calls: null. Which causes OpenAI assistant to raise exceptions due to the following check:

raise ArgumentError, "Tool calls must be an array of hashes" unless tool_calls.is_a?(Array) && tool_calls.all? { |tool_call| tool_call.is_a?(Hash) }

This fix is to make the check consider nil as an empty array. This change is related to #1011


littlechu:
issue seem to have fix with gem update for some of the require library. But I think the code should deal with the case where tool_calls being pass in as nil

PR #623, opened 2024-05-15
Adding GCP scope to auth in google_vertex_ai.rb

palladius:
adding scope as it often gives me error. See  issue #622 

(https://github.com/patterns-ai-core/langchainrb/issues/622)

Error being fixed:

```
 VertexLLM.authorizer.fetch_access_token
(irb):1:in `<main>': Authorization failed.  Server message: (Signet::AuthorizationError)
{"error":"invalid_scope","error_description":"Invalid OAuth scope or ID token audience provided."}
```

PR #1020, opened 2025-07-31
Fix incorrect usage of auth scope in Google::Auth.get_application_default

ytnk531:
Updated the method call to pass the auth scopes as an array inside the `options` hash, instead of directly as the `scope` argument. This ensures that authentication is configured correctly, per the expected interface: https://github.com/googleapis/google-auth-library-ruby/blob/googleauth/v1.14.0/lib/googleauth/application_default.rb#L53

Currently, due to this bug, the scopes are not passed correctly to GCECredentials. As a result, authentication from VM instances fails, preventing the use of Vertex AI.  
Merging this PR will fix this critical issue.


czlee:
I'm also seeing this problem, dug into it before I saw this PR, and came to the same conclusion: The `scope` argument here should be positional, not keyword.

The current effect is that credentials **must** be set from env or path. They cannot be left to [Application Default Credentials (ADC)](https://docs.cloud.google.com/docs/authentication/provide-credentials-adc). This fix would make credentials via ADC accessible.

The `scope` argument was added in #623. I wondered why this problem didn't surface at the time, so looked into it further. The following analysis is explains why this doesn't happen when configured from env vars _etc._, but only happens when falling back to ADC.

## Analysis

Langchain.rb calls `Google::Auth.get_application_default` in [langchain/llm/google_vertex_ai.rb#L28-L34](https://github.com/patterns-ai-core/langchainrb/blob/0.19.5/lib/langchain/llm/google_vertex_ai.rb#L28-L34):
```rb
      @authorizer = ::Google::Auth.get_application_default(scope: [
        "https://www.googleapis.com/auth/cloud-platform",
        "https://www.googleapis.com/auth/generative-language.retriever"
      ])
```

`Google::Auth.get_application_default` looks like this ([googleauth/application_default.rb#L53-L59](https://github.com/googleapis/google-auth-library-ruby/blob/googleauth/v1.14.0/lib/googleauth/application_default.rb#L53-L59)):
```rb
    def get_application_default scope = nil, options = {}
      creds = DefaultCredentials.from_env(scope, options) ||
              DefaultCredentials.from_well_known_path(scope, options) ||
              DefaultCredentials.from_system_default_path(scope, options)
      return creds unless creds.nil?
      raise NOT_FOUND_ERROR unless GCECredentials.on_gce? options
      GCECredentials.new options.merge(scope: scope)
```

The `DefaultCredentials` methods all route to `#interpret_options` ([googleauth/credentials_loader.rb#L178-L182](https://github.com/googleapis/google-auth-library-ruby/blob/1ef191b1342c19736ebe7a5640bfb86a7d989c68/lib/googleauth/credentials_loader.rb#L178-L185)). This is more permissive: Although it expects `scope` as a positional argument, if it is passed a hash with the `:scope` key, it will still work.
```rb
      def interpret_options scope, options
        if scope.is_a? Hash
          options = scope
          scope = nil
        end
        return options.merge scope: scope if scope && !options[:scope]
        options
      end
```

However, the `GCECredentials.new` expects an `options` hash, which `options.merge(scope: scope)` provides. However because Langchain.rb used a keyword argument, this becomes a nested hash, `scope: { scope: [...] }`, which is incorrect.

PR #1024, opened 2025-08-30
Fix anthropic system messages

noa-kaplan:
1. Add `system` role to `AnthropicMessage` `ROLES` constant
2. Implement proper `system?` method that returns `true` for system role
3. Add `system_hash` method for converting system messages to API format
4. Update `to_hash` method to handle system messages correctly
5. Fix `support_system_message?` to return `true` (was incorrectly returning false)
6. Update tests to expect correct behavior for system messages


andreibondarev:
@noa-kaplan Is the `role: "system"` getting added? I couldn't find any mention of it: https://docs.anthropic.com/en/api/messages

PR #1027, opened 2025-09-12
Rename Langchain.rb to LangChain.rb

koic:
In the LLM field, the name LangChain is widely recognized, and the project was likely originally intended to be called LangChain.rb as a product: https://github.com/patterns-ai-core/langchainrb/commit/ab9d9889c2183d99a90bee1347fe1fed0119e4c1

It appears that the current use of the name `Langchain` is due to implementation constraints. This PR updates the behavior so that Zeitwerk also recognizes the `LangChain` module name for the `langchain.rb` file.

Both the existing `Langchain` and the new `LangChain` module names will work. However, the `Langchain` constant is now marked as deprecated, and a warning will be shown when it is used:

```console
$ bundle exec ruby -Ilib -rrails -rlangchainrb -e 'Langchain::Errors'
-e:1: warning: `LangChain` is deprecated. Use `LangChain` instead.
```

This change makes it possible to use the name `LangChain`, which is the standard name in the AI ecosystem. Because `LangChain` is a proper noun, renaming to this widely recognized and commonly used name follows the principle of least surprise for users.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -25,7 +25,7 @@ If applicable, add screenshots to help explain your problem.
 
 - OS: [e.g. OS X, Linux, Ubuntu, Windows]
 - Ruby version [e.g. 3.1, 3.2, 3.3]
-- Langchain.rb version [e.g. 0.13.0]
+- LangChain.rb version [e.g. 0.13.0]
 
 **Additional context**
 Add any other context about the problem here.
--- a/.standard.yml
+++ b/.standard.yml
@@ -2,5 +2,5 @@ ignore:
   - "**/*":
     - Style/ArgumentsForwarding
 
-# Specify the minimum supported Ruby version supported by Langchain.rb.
+# Specify the minimum supported Ruby version supported by LangChain.rb.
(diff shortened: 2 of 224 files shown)


sandstrom:
Out of curiosity, what's the relationship between this gem/repo and the [original langchain](https://en.wikipedia.org/wiki/LangChain) project? 

Because that project is now driven by a for-profit startup, and this seems more like a true open-source project that isn't bound to a specified company.

But also, if this is independent, maybe there is also some risk of confusion?

***

Even more unrelated, is there any discussion of collaboration between this project and RubyLLM?

One thing that seems like a good candidate for collaboration between these two projects are the model-provider wrappers and/or model registry.

So basically providing a unified API that hides the small differences between different providers. Maybe there are other 'building block' parts too, that would make sense to share?

Of course each project would still be independent and have different ideas about public APIs, gem design and purpose/aim of each project.

cc @crmne

scarroll32:
> Out of curiosity, what's the relationship between this gem/repo and the [original langchain](https://en.wikipedia.org/wiki/LangChain) project?

That's a good question. I had come to this project assuming it was the (missing) Ruby extension for LangChain and associated products like LangSmith.

PR #1037, opened 2026-05-30
Document Tuning Engines with the OpenAI adapter

cerebrixos:
## Summary

- Adds a docs-only example for using Langchain.rb with Tuning Engines through the existing OpenAI-compatible `uri_base` configuration.
- Keeps Langchain.rb APIs and runtime behavior unchanged: no dependency, adapter, or code-path changes.
- Shows a concrete path for Ruby/Rails teams that want to keep Langchain.rb for prompts, chains, tools, and app logic while routing model calls through a governed OpenAI-compatible endpoint.

## Why this belongs here

Tuning Engines is an AI control plane for teams that want centralized model access, routing, tenant-scoped keys, policy/guardrail checks, audit logs, traces, approvals, and usage/cost visibility around existing AI application code.

Langchain.rb already supports this pattern via `Langchain::LLM::OpenAI` plus `llm_options[:uri_base]`. The proposed docs make that existing capability discoverable for Rails/Ruby teams without adding a new provider, new dependency, or new integration surface.

## Testing

- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,23 @@ llm = Langchain::LLM::OpenAI.new(
 )
 ```
 
+#### OpenAI-compatible gateways
+
+You can route the OpenAI LLM adapter through an OpenAI-compatible gateway by passing the underlying `ruby-openai` client options in `llm_options`. For example, Ruby and Rails apps using [Tuning Engines](https://www.tuningengines.com/) can keep the Langchain.rb interface while centralizing model routing, policy controls, audit logs, traces, approvals, and cost visibility:
+
+```ruby
+llm = Langchain::LLM::OpenAI.new(
+  api_key: ENV["TUNING_ENGINES_API_KEY"],
+  llm_options: {
+    uri_base: "https://api.tuningengines.com/v1"
+  },
+  default_options: {
+    chat_model: ENV.fetch("TUNING_ENGINES_MODEL", "gpt-4o-mini"),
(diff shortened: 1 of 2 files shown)


andreibondarev:
@cerebrixos what's tuning engines?

cerebrixos:
Thanks for asking. Tuning Engines is an AI control plane built for teams that already have app/agent code and want to route model calls through a governed OpenAI-compatible endpoint instead of wiring governance separately in each app.

The reason I thought this was relevant for Langchain.rb specifically is that Ruby/Rails teams can keep using `Langchain::LLM::OpenAI` and set the existing `llm_options[:uri_base]` to the Tuning Engines API. Langchain.rb still owns prompts, chains, tools, memory, and app logic; Tuning Engines sits behind the OpenAI-compatible interface for tenant-scoped keys, model routing, policy/guardrail checks, audit logs, traces, approvals, and usage/cost visibility.

So this PR is intentionally docs-only:

- no new dependency
- no adapter change
- no endorsement implied
- no runtime behavior change
- just an example of using the existing OpenAI adapter with a governed OpenAI-compatible endpoint

We also build with Ruby/Rails, so this is a real target audience for us: Rails teams that want governance and observability around Langchain.rb usage without rewriting their app around a separate SDK.

If you prefer, I can make the example more vendor-neutral and mention Tuning Engines only as one possible OpenAI-compatible endpoint.

andreibondarev:
@cerebrixos I don't think this belongs in the gem itself. This comment can live in the Tuning Engines docs itself.

PR #1040, opened 2026-08-08
Fix parallel_tool_calls: false for Gemini assistants

### LINKED ISSUE #1038: parallel_tool_calls: false does nothing when used with Gemini
Langchain::LLM::GoogleGemini and Langchain::LLM::GoogleVertexAI

Setting `parallel_tool_calls: false` when initializing Assistant is not having an effect. I still see the warning and face errors because it tries to call tools parallelly.

```
WARNING: `parallel_tool_calls:` is not supported by Google Gemini currently
```

```
StandardError: {"error" => {"code" => 400, "message" => "Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn.", "status" => "INVALID_ARGUMENT"}}
```

mikemikimike:
## Summary

Fixes #1038.

Google Gemini can return multiple function calls in one model response, but its function-calling configuration does not provide an OpenAI-style `parallel_tool_calls: false` switch. When an assistant was configured with `parallel_tool_calls: false`, Langchain.rb still submitted one tool response for every returned call in the same turn, which could leave Gemini's function-call/response parts out of sync.

This change makes the explicit `false` setting serialize tool execution at the model-turn boundary: the assistant keeps the first tool call on the model message, executes it, and lets the next model turn request the remaining calls. Existing `nil` and `true` behavior is unchanged.

## Validation

- `bundle exec rspec spec/lib/langchain/assistant/assistant_spec.rb --example 'serializes tool execution to one call per model turn' --format documentation` ? 1 example, 0 failures
- Google Gemini assistant context ? 13 examples, 0 failures
- OpenAI assistant context ? 30 examples, 0 failures
- Anthropic assistant context ? 13 examples, 0 failures
- Google Gemini adapter and message specs ? 6 examples, 0 failures
- `bundle exec standardrb` ? passed
- Ruby syntax checks and `git diff --check` ? passed

The full RSpec suite is currently blocked on this Windows environment because the existing `mistral-ai` dependency cannot load the native `libcurl` library while loading `spec/lib/langchain/llm/mistral_ai_spec.rb`; no examples ran in that full-suite invocation. The changed path and neighboring provider contexts pass independently.

### DIFF
--- a/lib/langchain/assistant.rb
+++ b/lib/langchain/assistant.rb
@@ -352,6 +352,15 @@ module Langchain
     #
     # @param tool_calls [Array<Hash>] The tool calls to run
     def run_tools(tool_calls)
+      if parallel_tool_calls == false
+        # Some providers, such as Google Gemini, do not expose an API option
+        # to disable parallel tool calls. Keep the model message in sync with
+        # the single tool response that we submit, then let the next turn
+        # request any remaining tool calls.
+        tool_calls = tool_calls.first(1)
+        messages.last.tool_calls = tool_calls
+      end
+
       # Iterate over each function invocation and submit tool output
       tool_calls.each do |tool_call|
         run_tool(tool_call)
--- a/lib/langchain/assistant/messages/base.rb
+++ b/lib/langchain/assistant/messages/base.rb
@@ -10,6 +10,8 @@ module Langchain
           :tool_calls,
           :tool_call_id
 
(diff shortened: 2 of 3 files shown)


andreibondarev:
@mikemikimike Thank you for the PR.

andreibondarev:
@mikemikimike Would you update the CHANGELOG please?

